PR #50, opened 2014-12-24
updating nr

e0d:


clintonb:
@e0d Any reason not to merge this?

e0d:
No, on phone or would merge now

PR #51, opened 2015-01-06
Added Endpoint for Course Problems

clintonb:
This replaces the submission counts endpoint. The latency from multiple calls to the submission counts kills performance on the answer distribution pages in Insights. This change creates a single endpoint that gets all problem data needed by Insights--submission counts and part IDs.

@dsjen @rocha @jbau @dylanrhodes

### DIFF
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -39,14 +39,15 @@ class ModelSerializerWithCreatedField(serializers.ModelSerializer):
     created = serializers.DateTimeField(format=settings.DATETIME_FORMAT)
 
 
-class ProblemSubmissionCountSerializer(serializers.Serializer):
+class ProblemSerializer(serializers.Serializer):
     """
-    Serializer for problem submission counts.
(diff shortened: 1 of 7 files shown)


rocha:
I think it makes sense to be able to also retrieve a problem individually, like we had previously: `/api/v0/courses/{course_id}/problems/{problem_id}`. Collecting all the problems seems to me to be more of an optimization, however there are also use cases in which we only require the data for one problem. @jbau is this compatible with the inline problem statistics?

clintonb:
@rocha It makes sense; but, it's not a use case we have, so not one I solved for.

rocha:
LGTM. :rocket: 

Maybe wait for @jbau or someone else from Stanford. I am unsure of how this relates to the inline analytics they are working on.

jbau:
I think this is fine.  Inline analytics figures out the course structure from LMS and doesn't really interact with the changed endpoints.

rocha:
Awesome, thanks for the clarification @jbau.

PR #52, opened 2015-01-09
Consolidation parameter for extraneous variants

dylanrhodes:
This PR adds logic and a boolean parameter to the answer distribution API which can consolidate rows for answers to questions which had randomization erroneously enabled.  This is a fix for the issue described in AN-4111 (<a href="https://openedx.atlassian.net/browse/AN-4111">here</a>) and in the thread <a href="https://github.com/edx/edx-analytics-pipeline/pull/40">here</a>.  There is an associated PR open against the API client to expose the consolidate parameter there <a href="https://github.com/edx/edx-analytics-data-api-client/pull/12">here</a>.  Feedback and comments encouraged!  @jbau @dcadams @sarina @clintonb

### DIFF
--- a/analytics_data_api/utils.py
+++ b/analytics_data_api/utils.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 from django.db.models import Q
 from rest_framework.authtoken.models import Token
 
@@ -31,3 +33,53 @@ def set_user_auth_token(user, key):
     Token.objects.create(user=user, key=key)
 
     print "Set API key for user %s to %s" % (user, key)
+
+
+def matching_tuple(answer):
+    """ Return tuple containing values which must match for consolidation. """
+    return (
+        answer.question_text,
+        answer.answer_value_text,
+        answer.answer_value_numeric,
+        answer.problem_display_name,
+        answer.correct,
+    )
+
+
+def consolidate_answers(problem):
+    """ Attempt to consolidate erroneously randomized answers. """
+    answer_sets = defaultdict(list)
+    match_tuple_sets = defaultdict(set)
+
+    for answer in problem:
+        answer.consolidated_variant = False
+
+        answer_sets[answer.value_id].append(answer)
+        match_tuple_sets[answer.value_id].add(matching_tuple(answer))
+
+    # If a part has more than one unique tuple of matching fields, do not consolidate.
+    for _, match_tuple_set in match_tuple_sets.iteritems():
+        if len(match_tuple_set) > 1:
+            return problem
+
+    consolidated_answers = []
+
+    for _, answers in answer_sets.iteritems():
+        consolidated_answer = None
+
+        if len(answers) == 1:
+            consolidated_answers.append(answers[0])
+            continue
+
+        for answer in answers:
+            if not consolidated_answer:
+                consolidated_answer = answer
+
+                consolidated_answer.variant = None
+                consolidated_answer.consolidated_variant = True
+            else:
+                consolidated_answer.count += answer.count
+
+        consolidated_answers.append(consolidated_answer)
+
+    return consolidated_answers
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -76,6 +76,29 @@ class ProblemResponseAnswerDistributionSerializer(ModelSerializerWithCreatedFiel
         )
 
 
+class ConsolidatedAnswerDistributionSerializer(ProblemResponseAnswerDistributionSerializer):
+    """
+    Serializer for consolidated answer distributions.
+    """
+
+    consolidated_variant = serializers.BooleanField()
+
+    class Meta(ProblemResponseAnswerDistributionSerializer.Meta):
+        fields = ProblemResponseAnswerDistributionSerializer.Meta.fields + ('consolidated_variant',)
+
+    # pylint: disable=super-on-old-class
+    def restore_object(self, attrs, instance=None):
+        """
+        Pops and restores non-model field.
+        """
+
+        consolidated_variant = attrs.pop('consolidated_variant', None)
+        distribution = super(ConsolidatedAnswerDistributionSerializer, self).restore_object(attrs, instance)
+        distribution.consolidated_variant = consolidated_variant
+
+        return distribution
+
+
 class GradeDistributionSerializer(ModelSerializerWithCreatedField):
     """
     Representation of the grade_distribution table without id
--- a/analytics_data_api/v0/views/problems.py
+++ b/analytics_data_api/v0/views/problems.py
@@ -1,11 +1,18 @@
+"""
+API methods for module level data.
+"""
+
+from itertools import groupby
+
 from rest_framework import generics
 
 from analytics_data_api.v0.models import ProblemResponseAnswerDistribution
-from analytics_data_api.v0.serializers import ProblemResponseAnswerDistributionSerializer
+from analytics_data_api.v0.serializers import ConsolidatedAnswerDistributionSerializer
 from analytics_data_api.v0.models import GradeDistribution
 from analytics_data_api.v0.serializers import GradeDistributionSerializer
 from analytics_data_api.v0.models import SequentialOpenDistribution
 from analytics_data_api.v0.serializers import SequentialOpenDistributionSerializer
+from analytics_data_api.utils import consolidate_answers
 
 
 class ProblemResponseAnswerDistributionView(generics.ListAPIView):
@@ -38,15 +45,30 @@ class ProblemResponseAnswerDistributionView(generics.ListAPIView):
             * variant: For randomized problems, the random seed used. If problem
               is not randomized, value is null.
             * created: The date the count was computed.
+
+    **Parameters**
+
+        You can request consolidation of response counts for erroneously randomized problems.
+
+        consolidate_variants -- If True, attempt to consolidate responses, otherwise, do not.
+
     """
 
-    serializer_class = ProblemResponseAnswerDistributionSerializer
+    serializer_class = ConsolidatedAnswerDistributionSerializer
     allow_empty = False
 
     def get_queryset(self):
         """Select all the answer distribution response having to do with this usage of the problem."""
         problem_id = self.kwargs.get('problem_id')
-        return ProblemResponseAnswerDistribution.objects.filter(module_id=problem_id)
+
+        queryset = ProblemResponseAnswerDistribution.objects.filter(module_id=problem_id).order_by('part_id')
+
+        consolidated_rows = []
+
+        for _, part in groupby(queryset, lambda x: x.part_id):
+            consolidated_rows += consolidate_answers(list(part))
+
+        return consolidated_rows
 
 
 class GradeDistributionView(generics.ListAPIView):
--- a/analytics_data_api/v0/tests/views/test_problems.py
+++ b/analytics_data_api/v0/tests/views/test_problems.py
@@ -20,23 +20,107 @@ class AnswerDistributionTests(TestCaseWithAuthentication):
     @classmethod
     def setUpClass(cls):
         cls.course_id = "org/num/run"
-        cls.module_id = "i4x://org/num/run/problem/RANDOMNUMBER"
-        cls.part_id1 = "i4x-org-num-run-problem-RANDOMNUMBER_2_1"
+        cls.module_id1 = "i4x://org/num/run/problem/RANDOMNUMBER"
+        cls.module_id2 = "i4x://org/num/run/problem/OTHERRANDOM"
+        cls.part_id = "i4x-org-num-run-problem-RANDOMNUMBER_2_1"
+        cls.correct = True
+        cls.value_id1 = '3'
+        cls.value_id2 = '4'
+        cls.answer_value_text = '3'
+        cls.answer_value_numeric = 3.0
+        cls.problem_display_name = 'Test Problem'
+        cls.question_text = 'Question Text'
+
         cls.ad1 = G(
             models.ProblemResponseAnswerDistribution,
             course_id=cls.course_id,
-            module_id=cls.module_id,
-            part_id=cls.part_id1
+            module_id=cls.module_id1,
+            part_id=cls.part_id,
+            correct=cls.correct,
+            value_id=cls.value_id1,
+            answer_value_text=cls.answer_value_text,
+            answer_value_numeric=cls.answer_value_numeric,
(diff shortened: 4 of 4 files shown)


edx-webhook:
Thanks for the pull request, @dylanrhodes! I've created [OSPR-344](https://openedx.atlassian.net/browse/OSPR-344) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads
- timeline information ('this must be merged by XX date', and why that is)
- partner information ('this is a course on edx.org')

All technical communication about the code itself will still be done via the Github pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

We can't start reviewing your pull request until you've added yourself to the [AUTHORS](https://github.com/edx/edx-analytics-data-api/blob/master/AUTHORS) file. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst) file for more information.

sarina:
Hey @dylanrhodes - help me a bit here with context, since I'm unfamiliar with Analytics work. Is this a new feature? Have you talked with the edX product team, and if not, do they need to be involved?

analytics_data_api/v0/views/problems.py
```diff
@@ -104,7 +112,22 @@ class ProblemResponseAnswerDistributionView(generics.ListAPIView):
     def get_queryset(self):
         """Select all the answer distribution response having to do with this usage of the problem."""
         problem_id = self.kwargs.get('problem_id')
-        return ProblemResponseAnswerDistribution.objects.filter(module_id=problem_id)
+        consolidate = self.request.QUERY_PARAMS.get('consolidate', 'False')
```
clintonb: It would be ideal if this parameter were not case-sensitive. This can be achieved by  calling `.lower()` on `consolidate` and comparing to `'true'` on the following line.

jbau: I also forgot to mention to @dylanrhodes earlier that we'd like this to be called `consolidate_variants`, to be a bit more verbose and clearer.

dylanrhodes: I've added both of your requested changes.

dylanrhodes:
Hey @sarina this is a fix for a problem in which instructors have been setting randomization to True for non-randomized questions as a workaround to get some other desired behavior (I think a check mark or no check mark or something).  That has been fixed, but there are still rows in the database which are near-identical duplicates (same answers to the same question) yet have different variant values because of the erroneous randomization setting.  This PR parameterizes the answer distribution API so that users can request consolidation of these near-duplicated rows for things like the inline analytics.  I have been in contact with the edX team through the discussions linked in the top PR comment above (AN-4111 and the earlier PR thread), and I think @jbau has spoken about it more recently with you guys.

analytics_data_api/v0/serializers.py
```diff
@@ -49,6 +49,26 @@ class ProblemSubmissionCountSerializer(serializers.Serializer):
     correct = serializers.IntegerField(default=0)
 
 
+class ConsolidatedAnswerDistributionSerializer(serializers.Serializer):
```
clintonb: You *may* be able to get away with subclassing `ProblemResponseAnswerDistributionSerializer` and just adding the `consolidated_variant` field.

analytics_data_api/utils.py
```diff
@@ -31,3 +31,38 @@ def set_user_auth_token(user, key):
     Token.objects.create(user=user, key=key)
 
     print "Set API key for user %s to %s" % (user, key)
+
+
+def matches(answer_one, answer_two):
+    """ Check whether two answers match sufficiently for consolidation. """
+    return (
```
clintonb: Parentheses are are not necessary here.

clintonb:
:+1:

analytics_data_api/v0/serializers.py
dylanrhodes: I've now made it a subclass of ProblemResponseAnswerDistributionSerializer, which makes its use a little more obvious but makes the declaration significantly more verbose.  Not totally sure which is better, but I think it's okay as a subclass.

analytics_data_api/utils.py
dylanrhodes: @clintonb Do you want me to switch to backslashes instead?  I think parentheses are <a href="https://www.python.org/dev/peps/pep-0008/#maximum-line-length">preferred</a> for multiline expressions in pep8, and I can't think of another way to express this multiline conditional without parentheses.

clintonb: Ah. In that case, disregard.

analytics_data_api/v0/serializers.py
```diff
@@ -75,6 +75,43 @@ class Meta(object):
...
 
+class ConsolidatedAnswerDistributionSerializer(ProblemResponseAnswerDistributionSerializer):
+    """
+    Serializer for consolidated answer distributions.
+    """
+
+    consolidated_variant = serializers.BooleanField()
+
+    class Meta(ProblemResponseAnswerDistributionSerializer.Meta):
+        fields = (
```
clintonb: I don't think this is necessary since you are subclassing the Meta class.

analytics_data_api/v0/serializers.py
```diff
@@ -75,6 +75,43 @@ class Meta(object):
...
+            'answer_value_numeric',
+            'problem_display_name',
+            'question_text',
+            'variant',
+            'consolidated_variant',
+            'created'
+        )
+
+    # pylint: disable=super-on-old-class
+    def restore_object(self, attrs, instance=None):
```
clintonb: Is this necessary for a read-only API?

clintonb:
@dylanrhodes I was thinking the serializer subclass might be as simple as subclassing and declaring the new field. If that's not the case, I am fine with reverting to the previous explicit definition to avoid the verbosity of subclassing.

analytics_data_api/v0/serializers.py
dylanrhodes: I can't access parent attributes in the class block, so I need to redeclare fields here to include consolidated_variants.

analytics_data_api/v0/serializers.py
dylanrhodes: This is the only way to add a non model field to a Django ModelSerializer at the moment.  There's an open request to add a non_native_fields field and an ExtensibleModelSerializer to the Django Rest Framework, but it isn't in our version.

analytics_data_api/v0/views/problems.py
```diff
@@ -104,7 +112,22 @@ class ProblemResponseAnswerDistributionView(generics.ListAPIView):
...
+        queryset = ProblemResponseAnswerDistribution.objects.filter(module_id=problem_id).order_by('part_id')
+
+        if not consolidate_variants:
+            return queryset
+
+        self.serializer_class = ConsolidatedAnswerDistributionSerializer
+        consolidated_rows = []
+
+        # pylint: disable=unused-variable
+        for part_id, part in groupby(queryset, lambda x: x.part_id):
```
jbau: you could do `for _, part in groupby(...)` instead to avoid the pylint directive.

analytics_data_api/v0/views/problems.py
```diff
@@ -104,7 +112,22 @@ class ProblemResponseAnswerDistributionView(generics.ListAPIView):
...
+            consolidated_rows += consolidate_answers([answer for answer in part])
```
jbau: also, `[answer for answer in part]` is pretty much a no-op, right?  or at least can't it be `list(part)` ?

analytics_data_api/utils.py
```diff
@@ -31,3 +31,38 @@ def set_user_auth_token(user, key):
...
+            answer_two = problem[j]
+
+            if matches(answer_one, answer_two):
+                answer_one.count += answer_two.count
+                answer_one.variant = None
+                answer_one.consolidated_variant = True
+
+                problem.remove(answer_two)
+            else:
+                j += 1
```
jbau: What about the case where there's one row that won't consolidate?  e.g.
```
part_id | variant | choice_id | answer text
  a     |  1      |  choice_1 | "foo"
  a     |  2      |  choice_1 | "foo"
  a     |  3      |  choice_1 | "bar"
  a     |  4      |  choice_1 | "baz"
 ```
In this case, I think we ought to be conservative and _not_ consolidate variants 1 and 2, since it's quite possible that the seed is taking effect but just mapping 1 and 2 to the same answer text value.  But this logic would consolidate those 2 cases.

I was thinking that consolidation would only happen if _all_ the rows of a particular `value_id` matched

Comment (jbau) [inline analytics_data_api/utils.py]: meaning for a part to be consolidated, it's necessary that for all of its `value_id`s, _all_ the rows with that `value_id` "match" in the other columns

File: analytics_data_api/v0/serializers.py (hunk 4 above)
Comment (jbau) [inline analytics_data_api/v0/serializers.py]: you could do `fields = ProblemResponseAnswerDistributionSerializer.Meta.fields + ('consolidated_variant',)` -- maybe that's nicer?

**Comment by shnayder** [comment]
I would really prefer it if we could agree on one behavior or the other, not have a flag that users of the API have to understand and developers have to test. For example, we could be conservative and merge only when we're really sure that things should be merged (e.g. when the same answer is present for many variants, with the same correctness value for each).  Thoughts?

**Comment by jbau** [comment]
I totally agree with being conservative (see my comments above).  If everyone can agree with the conditions for consolidation above, then perhaps that can just be always used, and we can eliminate the GET param?

File: analytics_data_api/v0/serializers.py (hunk 4 above)
Comment (dylanrhodes) [inline analytics_data_api/v0/serializers.py]: I switched it to the suggestion from @jbau - looks much nicer.

File: analytics_data_api/v0/views/problems.py (hunk 6 above)
Comment (dylanrhodes) [inline analytics_data_api/v0/views/problems.py]: Done

File: analytics_data_api/v0/views/problems.py (hunk 7 above)
Comment (dylanrhodes) [inline analytics_data_api/v0/views/problems.py]: Switched it out

File: analytics_data_api/utils.py (hunk 8 above)
Comment (dylanrhodes) [inline analytics_data_api/utils.py]: I've changed the logic so that answers in the same part are grouped by their value_id and then sets of answers sharing value_id are only consolidated if all members of the set match each other.  Do you also want to require that all of the sets satisfy that condition for any of them to be consolidated?

Comment (jbau) [inline analytics_data_api/utils.py]: Ah, nice catch.  I think so, because the behavior should be the same across all `value_id`s in a `part`

**Comment by dylanrhodes** [comment]
@clintonb The subclassed serializer is substantially less verbose including the suggestion from @jbau so I've left it in.  As to including the get parameter or not, it's optional in the current implementation, so while we do have to worry about it for testing, I don't think it's any additional cognitive load for users of the API or API client, since those who don't need to use it need not be aware of it or include it in their code.  I'm okay with taking it out, but it will add a bit of processing time to each API call if we're always checking if values can be consolidated before we return them.  The logic for consolidation is already pretty conservative (all columns except count, variant, and created must match for all rows sharing a given part_id and value_id).  @shnayder

File: analytics_data_api/utils.py [hunk 9]
```diff
@@ -31,3 +31,53 @@ def set_user_auth_token(user, key):
...
+        consolidate = len(answers) > 1
+
+        for i, answer in enumerate(answers):
+            j = i + 1
+
+            while consolidate and j < len(answers):
+                if not matches(answers[i], answers[j]):
+                    consolidate = False
+                else:
+                    j += 1
```
jbau: just recapping a hipchat discussion, I think this logic can be simplified by using python's `set` on tuples that are generated from `answers` like how `matches` does it now.

dylanrhodes:
@jbau @clintonb This is ready for another look.  At Jason's request, I made the requirements for consolidation stricter (now all value_id sets for a part must pass the check for consolidation - all rows match in all fields except count, created, and variant - for any to be consolidated).  The code in consolidate_answers is also a little cleaner.  Let me know if you have any other comments for it.  Thanks!

shnayder:
Are there times when you wouldn't want to consolidate? I'd want to always have it on in Insights, and imagine that you're adding it because you want it always on in your inline analytics view as well, right? Which means performance needs to be good enough for us to use it at all. My inclination would be to have it always on, and re-evaluate if it actually causes noticeable performance issues at some point, or someone has compelling use cases for making it optional.

jbau:
@shnayder that sounds good.  so basically what you're proposing is eliminating the GET param and just want having it consolidate on by default, all the time.

shnayder:
@jbau yup

dylanrhodes:
Hey everyone, I've responded to the most recent comments including removing the boolean parameter for consolidation in favor of making the logic mandatory.  The most recent revision has updated test cases to reflect the new state of affairs and seems ready to go. Feedback and comments are welcome! @jbau @shnayder @clintonb

analytics_data_api/v0/serializers.py
```diff
@@ -76,6 +76,29 @@ class Meta(object):
         )
 
 
+class ConsolidatedAnswerDistributionSerializer(ProblemResponseAnswerDistributionSerializer):
```
clintonb: `ProblemResponseAnswerDistributionSerializer` can be removed since it is no longer in use.

clintonb:
:+1: Looks good. I can go either way on `ProblemResponseAnswerDistributionSerializer`; but, I am more in favor of removing it as I cannot foresee a future use case for it.

jbau:
@clintonb since it serves as the base class of `ConsolidatedAnswerDistributionSerializer` I don't think it's outlandish to keep a distinct `ProblemResponseAnswerDistributionSerializer` around.

dylanrhodes:
@clintonb @shnayder @sarina @jbau Should I go ahead and merge this?

jbau:
:+1:

clintonb:
Merge away.

sarina:
You don't need my approval :) and traditionally, two thumbs is sufficient for merging, unless you absolutely need review/approval from additional people.

clintonb:
@dylanrhodes Do you have merge rights to this repo...or have I been holding up this process?

sarina:
@clintonb the safe expectation is that if someone doesn't work in this building they don't have merge rights.

dylanrhodes:
@clintonb I think I do have merge rights here and to the api-client repo; I was just asking out of an abundance of caution.  Measure twice, cut once, etc.

dsjen:
In the future, please tag me in PRs to this project too.

PR #53, opened 2015-01-21
Problem answer distribution will only return the answer_value.

dsjen:
@jbau @rocha @clintonb


jbau:
:+1:

clintonb:
:+1:

analytics_data_api/v0/views/problems.py
```diff
@@ -30,9 +30,7 @@ class ProblemResponseAnswerDistributionView(generics.ListAPIView):
             * count: The number of times the answer in this collection was
               given.
             * value_id: The ID of the answer in this collection.
-            * answer_value_text: The text of this answer, for text problems.
-            * answer_value_numeric: The number for this answer, for numeric
-              problems.
+            * answer_value: An answer for this problems.
```
brianhw: s/problems/problem

brianhw: Or use the description for answer_value_text.

dsjen: Good catch--updated to "problem."

PR #54, opened 2015-01-23
Optimized Courses Problems Endpoint

clintonb:
Using raw SQL to retrieve and aggregate the data. This significantly decreases response time. Also added a created field so that users know when the data was last updated.

@rocha @dsjen @jbau @mulby @brianhw @dylanrhodes

### DIFF
--- a/analytics_data_api/v0/views/courses.py
+++ b/analytics_data_api/v0/views/courses.py
@@ -4,6 +4,7 @@ import warnings
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import connections
 from django.db.models import Max
 from django.http import Http404
 from django.utils.timezone import make_aware, utc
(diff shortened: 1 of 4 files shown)


analytics_data_api/v0/views/courses.py
```diff
@@ -627,36 +629,31 @@ class ProblemsListView(BaseCourseView):
...
+GROUP BY module_id;
+        """
+        with connections[settings.ANALYTICS_DATABASE].cursor() as cursor:
+            cursor.execute(sql, [self.course_id])
+            rows = dictfetchall(cursor)
 
-        data = []
+        for row in rows:
+            # Convert the comma-separated list into an array of strings.
+            row['part_ids'] = row['part_ids'].split(',')
```
dsjen: The previous code had `part_ids = set()    # Use a set to remove duplicate values.`.  It seems this change may introduce duplicates.

clintonb: The SQL query  has a distinct modifier on the `part_ids` value returned.

dsjen: Excellent!

dsjen:
Looks like the build is failing.  Otherwise, looks good to me!

PR #55, opened 2015-02-06
Fixing MySQL GROUP_CONCAT() Issues

clintonb:
The group_concat_max_len value needs to be increased for some courses. This value is specific to MySQL (as opposed to SQLite run locally), so the DB vendor must be compared to ensure the new code only runs for MySQL databases.

@dsjen @jbau @dylanrhodes

### DIFF
--- a/analytics_data_api/v0/views/courses.py
+++ b/analytics_data_api/v0/views/courses.py
@@ -645,7 +645,14 @@ FROM answer_distribution
 WHERE course_id = %s
 GROUP BY module_id;
         """
-        with connections[settings.ANALYTICS_DATABASE].cursor() as cursor:
+        connection = connections[settings.ANALYTICS_DATABASE]
+        with connection.cursor() as cursor:
+            if connection.vendor == 'mysql':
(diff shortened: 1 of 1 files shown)


analytics_data_api/v0/views/courses.py
```diff
@@ -645,7 +645,14 @@ def get_queryset(self):
 WHERE course_id = %s
 GROUP BY module_id;
         """
-        with connections[settings.ANALYTICS_DATABASE].cursor() as cursor:
+        connection = connections[settings.ANALYTICS_DATABASE]
+        with connection.cursor() as cursor:
+            if connection.vendor == 'mysql':
+                # The default value of group_concat_max_len, 1024, is too low for some course data. Increase this value
+                # to it's maximum possible value. For more information see
```
dsjen: Nice comment!  "it's" should be "its" rather.

dsjen:
It looks like `max_packet_size` is 4194304 on stage at least.  Looks good to me!

@brianhw any concerns?

clintonb:
Still awaiting feedback here.

PR #58, opened 2015-02-11
Enable stats reporting to datadog.

benpatterson:
@clintonb @dsjen 

Got the bash script but didn't enable it! This will actually turn it on. We've made the adjustments on the Travis side so arguably this will turn on this kind of reporting.


dsjen:
:+1:

clintonb:
:+1:

PR #59, opened 2015-02-17
Returning HTTP 503 if service is unhealthy

clintonb:
@feanil @dsjen @rlucioni

### DIFF
--- a/analyticsdataserver/tests.py
+++ b/analyticsdataserver/tests.py
@@ -46,7 +46,7 @@ class OperationalEndpointsTest(TestCaseWithAuthentication):
     def test_health(self):
         self.assert_database_health('OK')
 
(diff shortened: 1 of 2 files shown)


analyticsdataserver/tests.py
```diff
@@ -46,7 +46,7 @@ def test_authentication_check_success(self):
     def test_health(self):
         self.assert_database_health('OK')
 
-    def assert_database_health(self, status):
+    def assert_database_health(self, status, status_code=200):
```
rlucioni: FYI, you can use constants from Django REST framework's `status` module instead of bare status codes. It's not a big deal, but it's a little more readable.

rlucioni:
:+1:

PR #63, opened 2015-02-27
Remove a SELECT 1 query

mulby:
This call to "exists()" causes a query like this to be run:

```sql
SELECT (1) AS `a` FROM `course_enrollment_gender_daily` WHERE `course_enrollment_gender_daily`.`course_id` = '<course_id>' LIMIT 1;
```

This query is surprisingly slow (in spite of the limit clause). This patch simply runs the actual queries we want to run and returns a 404 if no data is found.

Reviewer: @clintonb

### DIFF
--- a/analytics_data_api/v0/views/courses.py
+++ b/analytics_data_api/v0/views/courses.py
@@ -39,19 +39,14 @@ class BaseCourseView(generics.ListAPIView):
 
         return super(BaseCourseView, self).get(request, *args, **kwargs)
 
-    def verify_course_exists_or_404(self, course_id):
-        if self.model.objects.filter(course_id=course_id).exists():
-            return True
-
-        raise Http404
-
     def apply_date_filtering(self, queryset):
         raise NotImplementedError
 
     def get_queryset(self):
-        self.verify_course_exists_or_404(self.course_id)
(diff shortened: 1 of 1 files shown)


analytics_data_api/v0/views/courses.py
```diff
@@ -39,19 +39,14 @@ def get(self, request, *args, **kwargs):
...
-        raise Http404
-
     def apply_date_filtering(self, queryset):
         raise NotImplementedError
 
     def get_queryset(self):
-        self.verify_course_exists_or_404(self.course_id)
         queryset = self.model.objects.filter(course_id=self.course_id)
         queryset = self.apply_date_filtering(queryset)
+        if not queryset:
```
clintonb: This won't work as expected. `queryset` will be truthy at this point because it has not yet been evaluated. I whipped up an alternative on the branch named 404-alternative. Feel free to use it, if you'd like.

mulby:
Actually, hold off on this, it will definitely improve performance, but it's a behavior change, so we need to think about this in more depth.

We are going to evaluate performance during a load shortly and if we are still getting 500s at that time, we'll revisit this.

clintonb:
Okay.

PR #64, opened 2015-02-27
Remove check for course existence

mulby:
This looks good, we should rebase it onto the revision that is in production and push it to stage for testing.


clintonb:
I don't think we've made any backward-incompatible changes on master. We should be safe to push this to production.

clintonb:
Famous last words.

clintonb:
Gah! Branches. Moved to #66

PR #69, opened 2015-04-11
Use docker for builds.

benpatterson:


benpatterson:
@dsjen @clintonb

clintonb:
:+1:

PR #70, opened 2015-04-13
Removed milliseconds from created date in problems endpoint.

dsjen:


dsjen:
@jab5569 @mulby @brianhw -- Including the milliseconds was causing errors when loading the answer distribution fixture.

mulby:
was this never working?

dsjen:
At some point, loading the fixtures was working.  It must have broke when the `if not isinstance(created, datetime.datetime):` code was added in.

jab5569:
I verified these strftime commands as you have them in test_courses.py in the python command line. 

GTM. :+1:

PR #71, opened 2015-04-15
Added video and video timeline endpoints.

dsjen:
@jab5569 @brianhw @mulby 

Mocked Video Data:
```
[
  {
    "video_id": "0fac49ba",
    "encoded_module_id": "i4x-edX-DemoX-video-5c90cffecd9b48b188cbfea176bf7fe9",
    "duration": 500,
    "segment_length": 5,
    "start_views": 1234,
    "end_views": 510,
    "created": "2015-04-15T181146"
  }
]
```

Mocked Video Timeline:
```
[
  {
    "segment": 0,
    "num_users": 1096,
    "num_views": 1277,
    "created": "2015-04-15T181146"
  },
  {
    "segment": 1,
    "num_users": 897,
    "num_views": 80,
    "created": "2015-04-15T181146"
  },
  {
    "segment": 2,
    "num_users": 743,
    "num_views": 2466,
    "created": "2015-04-15T181146"
  },
 ...
```

### DIFF
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -241,3 +241,28 @@ class CourseActivityWeeklySerializer(serializers.ModelSerializer):
         model = models.CourseActivityWeekly
         # TODO: Add 'posted_forum' here to restore forum data
         fields = ('interval_start', 'interval_end', 'course_id', 'any', 'attempted_problem', 'played_video', 'created')
+
+
+class VideoSerializer(ModelSerializerWithCreatedField):
+    class Meta(object):
+        model = models.Video
+        fields = (
+            'pipeline_video_id',
+            'encoded_module_id',
+            'duration',
+            'segment_length',
+            'start_views',
+            'end_views',
+            'created'
+        )
+
+
+class VideoTimelineSerializer(ModelSerializerWithCreatedField):
+    class Meta(object):
+        model = models.VideoTimeline
+        fields = (
+            'segment',
+            'num_users',
+            'num_views',
+            'created'
+        )
--- a/analytics_data_api/management/commands/generate_fake_course_data.py
+++ b/analytics_data_api/management/commands/generate_fake_course_data.py
@@ -78,6 +78,7 @@ class Command(BaseCommand):
 
         # Delete existing data
         for model in [models.CourseEnrollmentDaily,
+                      models.CourseEnrollmentModeDaily,
                       models.CourseEnrollmentByGender,
                       models.CourseEnrollmentByEducation,
                       models.CourseEnrollmentByBirthYear,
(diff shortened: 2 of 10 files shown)


dsjen:
@jab5569 @brianhw @mulby 

Mocked Video Data:
```
[
  {
    "video_id": "0fac49ba",
    "encoded_module_id": "i4x-edX-DemoX-video-5c90cffecd9b48b188cbfea176bf7fe9",
    "duration": 500,
    "segment_length": 5,
    "start_views": 1234,
    "end_views": 510,
    "created": "2015-04-15T181146"
  }
]
```

Mocked Video Timeline:
```
[
  {
    "segment": 0,
    "num_users": 1096,
    "num_views": 1277,
    "created": "2015-04-15T181146"
  },
  {
    "segment": 1,
    "num_users": 897,
    "num_views": 80,
    "created": "2015-04-15T181146"
  },
  {
    "segment": 2,
    "num_users": 743,
    "num_views": 2466,
    "created": "2015-04-15T181146"
  },
 ...
```

analytics_data_api/v0/serializers.py
```diff
@@ -241,3 +241,31 @@ class Meta(object):
         model = models.CourseActivityWeekly
         # TODO: Add 'posted_forum' here to restore forum data
         fields = ('interval_start', 'interval_end', 'course_id', 'any', 'attempted_problem', 'played_video', 'created')
+
+
+class VideoSerializer(ModelSerializerWithCreatedField):
+    # pipeline_video_id is renamed to video_id
```
mulby: why? I'm also worried about "video_id" confusion here...

analytics_data_api/v0/serializers.py
```diff
@@ -241,3 +241,31 @@ class Meta(object):
...
+            'end_views',
+            'created'
+        )
+
+
+class VideoTimelineSerializer(ModelSerializerWithCreatedField):
+    class Meta(object):
+        model = models.VideoTimeline
+        fields = (
+            'segment',
```
mulby: it just occurred to me that in the data I generated "segment" is actually "seconds"... so instead of seeing 0,1,2,3... you will see 0,5,10,15...

We probably want the former?

mulby:
Generally looks good, my only blocking question is the one about "video_id" vs. "pipeline_video_id"

analytics_data_api/v0/serializers.py
dsjen: Now that you mention it, I think you're right.  I was thinking that if the video_id changed formats (maybe we find the edx video IDs are available), we could just leave the field as is and swap out the logic.

I do think it's probably better to be specific about where the ID is coming from though--less confusing!  I'll update the PR in a bit.

analytics_data_api/v0/serializers.py
dsjen: Ah, good catch.  I think the former seems to make more sense to me.  We're binning views into buckets/segments.  If we changed the field to seconds, it would seems to imply that we are counting the number of views and users at that particular discrete time point.

analytics_data_api/v0/serializers.py
dsjen: Updated.

mulby:
:+1:

PR #72, opened 2015-05-04
Update fixture for answer_distribution. 

brianhw:
Modify fixture data to include first_response_count and last_response_count instead of count.

Also change date on course data generation so it doesn't take as long.

@dsjen @mulby @jab5569


jab5569:
LGTM :+1:

dsjen:
Thank you, @brianhw!  :+1:

PR #75, opened 2015-05-21
Doc: update course video data response values

mhoeber:
@dsjen @lamagnifica

### DIFF
--- a/docs/api/source/change_log.rst
+++ b/docs/api/source/change_log.rst
@@ -8,5 +8,8 @@ Change Log
 
    * - Date
      - Change
-   * - 05/18/15
+   * - 21 May 2015
+     - Updated :ref:`Get the Course Video Data` to add the ``users_at_start``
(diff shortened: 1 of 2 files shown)


dsjen:
Looks great! :+1:

docs/api/source/change_log.rst
```diff
@@ -8,5 +8,8 @@ Change Log
 
    * - Date
      - Change
-   * - 05/18/15
+   * - 21 May 2015
+     - Updated :ref:`Get the Course Video Data` to add the ``users_at_start``
```
lamagnifica: add or rename? looks like a name change maybe?

lamagnifica:
1 possible clarification in the change log. :+1:

docs/api/source/change_log.rst
mhoeber: it's not really a rename - it's different fields

PR #77, opened 2015-06-04
Updated problem endpoint to return problem submissions.

dsjen:
* total_submissions and correct_submissions were returning problem part submissions previously

@jab5569 @mulby @brianhw


brianhw:
LGTM. :+1:

jab5569:
:+1: LGTM

PR #78, opened 2015-06-22
Added cumulative_count field to enrollment/mode endpoint.

dsjen:
This assumes that the `cumulative_count` column is in the `course_enrollment_mode_daily` table.  If the data pipeline is ready to generate this data, I can go ahead and merge this in once reviewed.

@jab5569 @mulby @brianhw

### DIFF
--- a/analytics_data_api/management/commands/generate_fake_course_data.py
+++ b/analytics_data_api/management/commands/generate_fake_course_data.py
@@ -91,6 +91,7 @@ class Command(BaseCommand):
         # Create new data
         daily_total = 1500
         date = start_date
+        cumulative_count = 0
 
         while date <= end_date:
             daily_total = get_count(daily_total)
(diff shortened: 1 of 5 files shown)


brianhw:
LGTM. :+1:

analytics_data_api/management/commands/generate_fake_course_data.py
```diff
@@ -98,8 +99,9 @@ def generate_daily_data(self, course_id, start_date, end_date):
 
             for mode, ratio in enrollment_mode_ratios.iteritems():
                 count = int(ratio * daily_total)
+                cumulative_count = max(cumulative_count + 10, count)
```
jab5569: Why +10?

jab5569:
Just one question, otherwise :+1:

analytics_data_api/management/commands/generate_fake_course_data.py
dsjen: My thought was that the cumulative counts will generally go higher as the counts change.  It's not particularly realistic, but does generate the data.

mulby:
:+1: 

Side note - we should discuss ways to tell people what versions of edx-analytics-pipeline, edx-analytics-data-api and edx-analytics-dashboard can be run together. Maybe we should join the Open edX named releases?

If someone deployed this change without upgrading their pipeline, their deployment would not be functional.

mamigot:
Hey @mulby , while the documentation is written, is there any way to know which versions of of edx-analytics-pipeline, edx-analytics-data-api and edx-analytics-dashboard go with each other? Thank you!!

mulby:
@mamigot Unfortunately we don't yet have a good way of indicating this. If you use the latest release tag from each repo it should work.

mamigot:
Thank you, @mulby , I will!

PR #82, opened 2015-08-23
API for retrieving # of passing students vs. time

bradenmacdonald:
**Description**: This adds a new API endpoint to the analytics API server for retrieving data from the [Grades Import Pipeline](https://github.com/edx/edx-analytics-pipeline/pull/149). Specifically, the new endpoint gives the number and percentage of students who are passing the given course as a function of time.

**Dependencies**: 
- https://github.com/edx/edx-analytics-pipeline/pull/149 (Grades Import Pipeline)

**Screenshot**:
Here's a screenshot of Swagger, which gives a nice overview of the new API endpoint:
screen shot 2015-08-22 at 11 41 24 pm

**Partner information**: not an edX partner - 3rd party-hosted open edX instance

**Reviewers**: TBD

### DIFF
--- a/analytics_data_api/v0/models.py
+++ b/analytics_data_api/v0/models.py
@@ -148,8 +148,23 @@ class CourseEnrollmentByCountry(BaseCourseEnrollment):
         unique_together = [('course_id', 'date', 'country_code')]
 
 
+class CourseGradeBreakdown(models.Model):
+    """ The number of students who have a given letter grade for a given course and date """
+
+    class Meta(object):
+        db_table = 'grade_breakdown'
+
+    course_id = models.CharField(db_index=True, max_length=255)
+    date = models.DateField(null=False)
+    letter_grade = models.CharField(max_length=64, null=True)
+    num_students = models.IntegerField(null=False)
+    percent_students = models.FloatField()
+    is_passing = models.BooleanField(null=False)
+    created = models.DateTimeField(auto_now_add=True)
+
+
 class GradeDistribution(models.Model):
-    """ Each row stores the count of a particular grade on a module for a given course. """
+    """ Each row stores the count of a particular score on a module for a given course. """
 
     class Meta(object):
         db_table = 'grade_distribution'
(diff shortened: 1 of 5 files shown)


analytics_data_api/v0/models.py
```diff
@@ -148,8 +148,23 @@ class Meta(BaseCourseEnrollment.Meta):
...
+    letter_grade = models.CharField(max_length=64, null=True)
+    num_students = models.IntegerField(null=False)
+    percent_students = models.FloatField()
+    is_passing = models.BooleanField(null=False)
+    created = models.DateTimeField(auto_now_add=True)
+
+
 class GradeDistribution(models.Model):
-    """ Each row stores the count of a particular grade on a module for a given course. """
+    """ Each row stores the count of a particular score on a module for a given course. """
```
itsjeyd: @bradenmacdonald If this model represents a score distribution, maybe the terminology should be changed in the rest of the class as well? Not sure if you would want to do that as part of this PR (it would probably exceed the scope of the PR a bit), but I wanted to point it out just in case.

itsjeyd:
@bradenmacdonald Looking good overall, I just have one question: Any specific reason why there is no endpoint (and serializer) for the `StudentGrade` model? We'll need `percent_grade` and `is_passing` info for OC-828, and I think it would make sense to include an endpoint for that in this PR -- that way we wouldn't have to open a separate data API PR for OC-828.

bradenmacdonald:
@itsjeyd Good question. The main reason I didn't put the StudentGrade endpoint in yet is that I was waiting for the user profiles PR to merge, so this PR can add that into "users.py". I'm also undecided if the grade should be a property of the course ("GET /course/:user_id/student_grade/:student_id") or the user ("GET /user/:user_id/course_grade/:student_id") - or maybe both? Do you have any thoughts?

bradenmacdonald:
@itsjeyd I looked into adding `StudentGrade` model and wrote some code: bdba9ae627c70fba4fa8598246f2dd6dd53ad16d. The issue is that it depends on the user profiles PR, so I think I'm going to wait until that PR is merged, then I'll rebase this branch on top and add in the StudentGrade endpoint. If upstream reviews the PRs in order, that should still let us avoid needing an extra PR.

itsjeyd:
@bradenmacdonald Had a look at the WIP code, it looks good. Making per-student info available via `users/{username}/grades/` is a good approach I think. It directly supports the use case that we'll implement for OC-828 and as far as know we don't really need an endpoint that is both course and user-specific right now.

PR #85, opened 2015-09-24
Add label for inter sphinx

mhoeber:
And Open edX name

@lamagnifica


lamagnifica:
?
 looks fine, are all of the APIs open? my understanding was that this could not be used externally yet, but don't know the philosophy behind the decision (or oversight)

mhoeber:
yes, I believe we should be making this change. it's really only available for people to use on their own open edx instance; outsiders can't use for edx.org data

lamagnifica:
:+1:

PR #86, opened 2015-11-16
Added settings for elasticsearch.

dsjen:
@dan-f -- this isn't one of the more exciting PRs.

I've added configurations for setting up elastic search.  When it comes time to implement, we can do something like:

```
from elasticsearch_dsl import connections, Index, DocType, String

connections.connections.create_connection(hosts=[settings.ELASTIC_SEARCH_LEARNERS_HOST)
roster = Index(settings.ELASTIC_SEARCH_LEARNERS_INDEX)

# this decorator sets up the index and we could move that step into Meta if this is overkill
@roster.doc_type
class RosterEntry(DocType):
     username = String(index='not_analyzed')
     class Meta:
          doc_type = 'roster_entry'  # this was a key part

# execute search
RosterEntry.search().execute()
```

### DIFF
--- a/analyticsdataserver/settings/base.py
+++ b/analyticsdataserver/settings/base.py
@@ -1,5 +1,6 @@
 """Common settings and globals."""
 
+from os import environ
 from os.path import abspath, basename, dirname, join, normpath
 from sys import stderr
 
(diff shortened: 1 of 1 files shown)


analyticsdataserver/settings/base.py
```diff
@@ -49,6 +49,11 @@
     }
 }
 ########## END DATABASE CONFIGURATION
+ELASTIC_SEARCH_LEARNERS_HOST = None
```
dan-f: Nit: I'd remove the first underscore in each name to be: `"ELASTICSEARCH_LEARNERS_HOST"`.

dan-f:
:+1: other than my nitpick.

dan-f:
Actually - could we make this variable be settable from the environment?

analyticsdataserver/settings/base.py
dsjen: Ah!  I meant for it not to have an underscore.  Updating.

dsjen:
@dan-f -- what about something like: 
```
ELASTICSEARCH_LEARNERS_HOST = os.environ.get('ELASTICSEARCH_LEARNERS_HOST', None)
```

dsjen:
I updated the config to use environment variables or `None`.

PR #87, opened 2015-11-18
Added learner details endpoint.

dsjen:
@mulby, @dan-f -- I've added the learner details endpoint.  This PR incorporates elastic search and a stub for the learners list endpoint too.

There are a few entity and event fields that are a bit inconsistent between various specs and the real data, but they should be easy to update when we get aligned on the names.

Also, the path has updated here instead of `/api/learner_analytics/v0/learners`, it is now `/api/v0/learners` for both consistency and potential changes to the API recommendation.


dsjen:
Closing and rebasing off of feature branch.

PR #88, opened 2015-11-30
Added learner details endpoint.

dsjen:
@mulby, @dan-f -- I've added the learner details endpoint. This PR incorporates elastic search and a stub for the learners list endpoint too.

There are a few entity and event fields that are a bit inconsistent between various specs and the real data, but they should be easy to update when we get aligned on the names.

Also, the path has updated here instead of /api/learner_analytics/v0/learners, it is now /api/v0/learners for both consistency and potential changes to the API recommendation.

### DIFF
--- a/analytics_data_api/v0/apps.py
+++ b/analytics_data_api/v0/apps.py
@@ -1,0 +1,13 @@
+from django.apps import AppConfig
+from django.conf import settings
+from elasticsearch_dsl import connections
+
+
+class ApiAppConfig(AppConfig):
+
+    name = 'analytics_data_api.v0'
+
+    def ready(self):
+        super(ApiAppConfig, self).ready()
+        if settings.ELASTICSEARCH_LEARNERS_HOST:
+            connections.connections.create_connection(hosts=[settings.ELASTICSEARCH_LEARNERS_HOST])
--- a/analytics_data_api/v0/exceptions.py
+++ b/analytics_data_api/v0/exceptions.py
@@ -1,0 +1,52 @@
+import abc
+
+
+class BaseError(Exception):
+    """
+    Base error.
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    message = None
+
+    def __str__(self):
+        return self.message
+
+
+class LearnerNotFoundError(BaseError):
+    """
+    Raise learner not found for a course.
+    """
+    def __init__(self, *args, **kwargs):
+        course_id = kwargs.pop('course_id')
+        username = kwargs.pop('username')
+        super(LearnerNotFoundError, self).__init__(*args, **kwargs)
+        self.message = self.message_template.format(username=username, course_id=course_id)
+
+    @property
+    def message_template(self):
+        return 'Learner {username} not found for course {course_id}.'
+
+
+class CourseNotSpecifiedError(BaseError):
+    """
+    Raise if course not specified.
+    """
+    def __init__(self, *args, **kwargs):
+        super(CourseNotSpecifiedError, self).__init__(*args, **kwargs)
+        self.message = 'Course id/key not specified.'
+
+
+class CourseKeyMalformedError(BaseError):
+    """
+    Raise if course id/key malformed.
+    """
+    def __init__(self, *args, **kwargs):
+        course_id = kwargs.pop('course_id')
+        super(CourseKeyMalformedError, self).__init__(*args, **kwargs)
+        self.message = self.message_template.format(course_id=course_id)
+
+    @property
+    def message_template(self):
+        return 'Course id/key {course_id} malformed.'
--- a/analytics_data_api/v0/models.py
+++ b/analytics_data_api/v0/models.py
@@ -1,4 +1,6 @@
+from django.conf import settings
 from django.db import models
+from elasticsearch_dsl import DocType
 
 from analytics_data_api.constants import country, genders
 
@@ -206,3 +208,15 @@ class Video(BaseVideo):
 
     class Meta(BaseVideo.Meta):
         db_table = 'video'
+
+
+class RosterEntry(DocType):
+    # pylint: disable=old-style-class
+    class Meta:
+        index = settings.ELASTICSEARCH_LEARNERS_INDEX
+        doc_type = 'roster_entry'
+
+    @classmethod
+    def get_course_user(cls, course_id, username):
+        return cls.search().query('term', course_id=course_id).query(
+            'term', username=username).execute()
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -1,7 +1,12 @@
+from urlparse import urljoin
 from django.conf import settings
 from rest_framework import serializers
 
-from analytics_data_api.constants import enrollment_modes, genders
+from analytics_data_api.constants import (
+    engagement_entity_types,
+    engagement_events,
+    enrollment_modes,
+    genders,)
 from analytics_data_api.v0 import models
 
 
@@ -306,3 +311,35 @@ class VideoTimelineSerializer(ModelSerializerWithCreatedField):
             'num_views',
             'created'
         )
+
+
+class LearnerSerializer(serializers.Serializer):
+    username = serializers.CharField()
+    enrollment_mode = serializers.CharField()
+    name = serializers.CharField()
+    account_url = serializers.SerializerMethodField('get_account_url')
+    email = serializers.CharField()
+    segments = serializers.Field(source='segments')
+    engagements = serializers.SerializerMethodField('get_engagements')
+
+    # TODO: add these back in when the index returns them
+    # enrollment_date = serializers.DateField(format=settings.DATE_FORMAT, allow_empty=True)
+    # last_updated = serializers.DateField(format=settings.DATE_FORMAT)
+    # cohort = serializers.CharField(allow_none=True)
+
+    def get_account_url(self, obj):
+        if settings.LMS_USER_ACCOUNT_BASE_URL:
+            return urljoin(settings.LMS_USER_ACCOUNT_BASE_URL, obj.username)
+        else:
+            return None
+
+    def get_engagements(self, obj):
+        """
+        Add the engagement totals.
+        """
+        engagements = {}
+        for entity_type in engagement_entity_types.ALL:
+            for event in engagement_events.EVENTS[entity_type]:
+                metric = '{0}_{1}'.format(entity_type, event)
+                engagements[metric] = getattr(obj, metric, 0)
+        return engagements
--- a/analytics_data_api/v0/tests/views/test_learners.py
+++ b/analytics_data_api/v0/tests/views/test_learners.py
@@ -1,0 +1,108 @@
+import json
+
+from elasticsearch_dsl.connections import connections
+from mock import patch, Mock
+from rest_framework import status
+
+from analyticsdataserver.tests import TestCaseWithAuthentication
+
+
+class LearnerTests(TestCaseWithAuthentication):
+    path_template = '/api/v0/learners/{}/?course_id={}'
+
+    @classmethod
+    def get_fixture(cls):
+        return {
+            "took": 11,
+            "_shards": {
+                "total": 10,
+                "successful": 10,
+                "failed": 0
+            },
+            "hits": {
+                "total": 33701,
+                "max_score": 7.201823,
+                "hits": [{
+                    "_index": "roster_1_1",
+                    "_type": "roster_entry",
+                    "_id": "edX/DemoX/Demo_Course|ed_xavier",
+                    "_score": 7.201823,
+                    "_source": {
+                        "username": "ed_xavier",
+                        "enrollment_mode": "honor",
+                        "problems_attempted": 43,
+                        "problems_completed": 3,
+                        "problem_attempts": 0,
+                        "course_id": "edX/DemoX/Demo_Course",
+                        "videos_viewed": 6,
+                        "name": "Edward Xavier",
+                        "segments": ["has_potential"],
+                        "email": "ed_xavier@example.com"
+                    }
+                }]
+            }
+        }
+
+    @classmethod
+    def setUpClass(cls):
+        # mock the elastic search client
+        client = Mock()
+        client.search.return_value = cls.get_fixture()
+        connections.add_connection('default', client)
+
+    def test_get_user(self):
+        user_name = 'ed_xavier'
+        course_id = 'edX/DemoX/Demo_Course'
+        response = self.authenticated_get(self.path_template.format(user_name, course_id))
+        self.assertEquals(response.status_code, 200)
+
+        expected = {
+            "username": "ed_xavier",
+            "enrollment_mode": "honor",
+            "name": "Edward Xavier",
+            "email": "ed_xavier@example.com",
+            "account_url": "http://lms-host/ed_xavier",
+            "segments": ["has_potential"],
+            "engagements": {
+                "problem_attempts": 0,
+                "problems_attempted": 43,
+                "problems_completed": 3,
+                "videos_viewed": 6,
+                "discussions_contributed": 0
+            },
+        }
+        self.assertDictEqual(expected, response.data)
+
+    @patch('analytics_data_api.v0.models.RosterEntry.get_course_user', Mock(return_value=[]))
+    def test_not_found(self):
+        user_name = 'a_user'
+        course_id = 'edX/DemoX/Demo_Course'
+        response = self.authenticated_get(self.path_template.format(user_name, course_id))
+        self.assertEquals(response.status_code, status.HTTP_404_NOT_FOUND)
+        expected = {
+            u"error_code": u"no_learner_for_course",
+            u"developer_message": u"Learner a_user not found for course edX/DemoX/Demo_Course."
+        }
+        self.assertDictEqual(json.loads(response.content), expected)
+
+    def test_no_course_id(self):
+        base_path = '/api/v0/learners/{}'
+        path = (base_path).format('ed_xavier')
+        response = self.authenticated_get(path)
+        self.assertEquals(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+        expected = {
+            u"error_code": u"course_not_specified",
+            u"developer_message": u"Course id/key not specified."
+        }
+        self.assertDictEqual(json.loads(response.content), expected)
+
+    def test_bad_course_id(self):
+        path = self.path_template.format('ed_xavier', 'malformed-course-id')
+        response = self.authenticated_get(path)
+        self.assertEquals(response.status_code, status.HTTP_400_BAD_REQUEST)
+        expected = {
+            u"error_code": u"course_key_malformed",
+            u"developer_message": u"Course id/key malformed-course-id malformed."
+        }
+        self.assertDictEqual(json.loads(response.content), expected)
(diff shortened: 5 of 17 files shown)


analytics_data_api/v0/serializers.py
```diff
@@ -306,3 +310,30 @@ class Meta(object):
...
+
+    def get_engagement(self, obj):
+        """
+        Add the engagement totals.
+        """
+        engagement = {}
+        for entity_type in engagement_entity_types.ALL:
+            for event in engagement_events.EVENTS[entity_type]:
+                metric = '{0}_{1}'.format(entity_type, event)
+                print metric
```
dan-f: Forgot to remove a couple prints in here

analytics_data_api/v0/views/learners.py
```diff
@@ -0,0 +1,113 @@
...
+from analytics_data_api.v0.models import RosterEntry
+from analytics_data_api.v0.serializers import LearnerSerializer
+
+
+class CourseViewMixin(object):
+    """
+    Captures the course_id query arg and validates it.
+    """
+
+    COURSE_ID_PATTERN = r'(?P<course_id>[^/+]+[/+][^/+]+[/+][^/]+)'
```
dan-f: This is also defined in `analytics_data_api/v0/urls/courses.py`. Could we share it?

analytics_data_api/v0/views/learners.py
```diff
@@ -0,0 +1,113 @@
...
+    def get(self, request, *args, **kwargs):
+        self.course_id = request.QUERY_PARAMS.get('course_id', None)
+        if not self.course_id:
+            raise ParseError('Course ID was not specified')
+        if not re.search(self.COURSE_ID_PATTERN, self.course_id):
+            raise ParseError('Course ID malformed')
+        return super(CourseViewMixin, self).get(request, *args, **kwargs)
+
+
+class LearnersListView(CourseViewMixin, generics.ListAPIView):
```
dan-f: This is part of the next story, right?

analytics_data_api/v0/views/learners.py
```diff
@@ -0,0 +1,113 @@
...
+class CourseViewMixin(object):
+    """
+    Captures the course_id query arg and validates it.
+    """
+
+    COURSE_ID_PATTERN = r'(?P<course_id>[^/+]+[/+][^/+]+[/+][^/]+)'
+    course_id = None
+
+    def get(self, request, *args, **kwargs):
+        self.course_id = request.QUERY_PARAMS.get('course_id', None)
```
dan-f: We should use the `opaque_keys` library to validate the `course_id`.

analytics_data_api/v0/views/learners.py
```diff
@@ -0,0 +1,113 @@
...
+        return super(LearnerView, self).get(request, *args, **kwargs)
+
+    def get_queryset(self):
+        return RosterEntry.get_course_user(self.course_id, self.username)
+
+    def get_object(self, queryset=None):
+        queryset = self.get_queryset()
+        if len(queryset) == 1:
+            return queryset[0]
+        raise Http404("Object does not exist")
```
dan-f: Check out the error format [here](https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions#edXRESTAPIConventions-5.Errors) - we should make sure we're following that. I'm not sure how Http404 is treated within DRF.

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -0,0 +1,90 @@
...
+                "video_played": 6,
+                "forum_created": 0
+            }
+        }
+        from pprint import pprint
+        pprint(response.data)
+        self.assertDictEqual(expected, response.data)
+
+    @patch('analytics_data_api.v0.models.RosterEntry.get_course_user', Mock(return_value=[]))
+    def test_not_found(self):
```
dan-f: 404s happen when the username/course_id pair aren't found. We should try this with both an invalid username and an invalid course_id. Perhaps use `ddt` here?

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -0,0 +1,90 @@
...
+        self.assertEquals(response.status_code, 200)
+
+        expected = {
+            "username": "ed_xavier",
+            "enrollment_mode": "honor",
+            "name": "Edward Xavier",
+            "email": "ed_xavier@example.com",
+            "segments": ["has_potential"],
+            "engagement": {
+                "forum_commented": 0,
```
dan-f: These aren't the keys in the API doc.

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -0,0 +1,90 @@
...
+        response = self.authenticated_get(self.path_template.format(user_name, course_id))
+        self.assertEquals(response.status_code, 200)
+
+        expected = {
+            "username": "ed_xavier",
+            "enrollment_mode": "honor",
+            "name": "Edward Xavier",
+            "email": "ed_xavier@example.com",
+            "segments": ["has_potential"],
+            "engagement": {
```
dan-f: The API doc chooses they key "engagements" rather than "engagement". I don't really mind either way, but we should be consistent.

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -0,0 +1,90 @@
...
+        connections.add_connection('default', client)
+
+    def test_get_user(self):
+        user_name = 'ed_xavier'
+        course_id = 'edX/DemoX/Demo_Course'
+        response = self.authenticated_get(self.path_template.format(user_name, course_id))
+        self.assertEquals(response.status_code, 200)
+
+        expected = {
+            "username": "ed_xavier",
```
dan-f: We need to add "account_url"

dan-f:
Done with my first pass, @dsjen. I'm still not convinced about the old `api/v0/...` url structure, as it boxes us in. I looked a bit at the URL routing code. Due to the separation of the `analyticsserver` and `analytics_data_api` apps, we should be able to include another `learner_analytics` app in the same way that we include the `analytics_data_api` or `v0` apps.

https://github.com/edx/edx-analytics-data-api/blob/master/analyticsdataserver/urls.py#L15
https://github.com/edx/edx-analytics-data-api/blob/master/analytics_data_api/urls.py#L6

dsjen:
Thanks, @dan-f!  I had the URLs routed and had to add a bit of extra code at https://github.com/edx/edx-analytics-data-api/compare/dsjen/learner-engagement-timeline?expand=1#diff-1b6a5ff34357efc621dd247bb536cb3eR4, https://github.com/edx/edx-analytics-data-api/compare/dsjen/learner-engagement-timeline?expand=1#diff-110b488f2b07b14fdb4a2589e7da9e18R9, and https://github.com/edx/edx-analytics-data-api/compare/dsjen/learner-engagement-timeline?expand=1#diff-92cd74a9de7813a9d844d93410cf5cf3R198.  There's a bit of throw away code involved because we're supporting two conventions.

I wanted to provide a few links for reference and lets chat more.  I'll address your other comments in the meantime.

analytics_data_api/v0/serializers.py
dsjen: Thanks!  Updated.

analytics_data_api/v0/views/learners.py
dsjen: Oops!  I'll remove the learn list view and route.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: Ah!  I'll update to "engagements" instead.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: I pulled these out from the the "possible events" in https://openedx.atlassian.net/wiki/display/AN/Learner+Analytics.  Also, the naming isn't quite consistent between what's in the documentation, api spec, and actual index though.  On the API: spec we have:
* discussions_contributed
* problems_attempted
* problems_completed
* videos_viewed
* videos_completed

The table spec has:
* problem_attempted
* problem_completed
* video_played
* forum_created
* forum_responded
* forum_commented

The current index payload has:
* problems_attempted
* problems_completed
* discussion_activity
* videos_watched

@dan-f & @mulby -- does anyone have opinions on what the names should be?

analytics_data_api/v0/views/learners.py
dsjen: Good idea!  Updated.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: Do you mean that we throw a 404 when the username and course_id is malformed or simply that it they don't exist?

analytics_data_api/v0/tests/views/test_learners.py
dsjen: Added LMS_USER_ACCOUNT_BASE_URL to settings and updated serializer to return account_url field.

analytics_data_api/v0/views/learners.py
dsjen: Thanks, I hadn't seen that before.  Do you have an example of how that's implemented in the LMS?

analytics_data_api/v0/models.py
```diff
@@ -206,3 +208,22 @@ class Video(BaseVideo):
 
     class Meta(BaseVideo.Meta):
         db_table = 'video'
+
+
+if settings.ELASTICSEARCH_LEARNERS_HOST:
+    connections.connections.create_connection(hosts=[settings.ELASTICSEARCH_LEARNERS_HOST])
```
mulby: I think we should avoid executing logic like this on import.

I feel like this should really go in a "one-time startup" function like the one described here: https://docs.djangoproject.com/en/1.8/ref/applications/#django.apps.AppConfig.ready

I know we hacked together a "startup" function for the LMS, but since then I think Django has added first class support for that type of pattern. I assume it's the method I linked to above, but am not really sure, maybe worth talking to someone who is familiar with the newer versions of Django.

analytics_data_api/v0/models.py
```diff
@@ -206,3 +208,22 @@ class Video(BaseVideo):
...
+
+# By creating the roster index, we can use the @roster.doc_type decorator below
```
mulby: This comment implies that the index is actually created here, which I think is a little confusing.

analytics_data_api/v0/models.py
```diff
@@ -206,3 +208,22 @@ class Video(BaseVideo):
...
+
+@roster.doc_type
+class RosterEntry(DocType):
+    # pylint: disable=old-style-class
+    class Meta:
+        doc_type = 'roster_entry'
+
+    @classmethod
+    def get_course_user(cls, course_id, username):
+        return cls.search().filter('term', course_id=course_id).query(
```
mulby: Not too familiar with the syntax here, but I profiled the following query:

```json
{
    "size": 50,
    "from": 0,
    "query": {
        "bool": {
            "must": [
                {
                    "term": {
                        "roster_entry.username": "foo"
                    }
                },
                {
                    "term": {
                        "roster_entry.course_id": "bar"
                    }
                }
            ]
        }
    }
}
```

analytics_data_api/v0/models.py
```diff
@@ -206,3 +208,22 @@ class Video(BaseVideo):
...
+
+if settings.ELASTICSEARCH_LEARNERS_HOST:
+    connections.connections.create_connection(hosts=[settings.ELASTICSEARCH_LEARNERS_HOST])
+
+# By creating the roster index, we can use the @roster.doc_type decorator below
+roster = Index(settings.ELASTICSEARCH_LEARNERS_INDEX)
+
+
+@roster.doc_type
+class RosterEntry(DocType):
```
mulby: I'm a little surprised you don't have to declare the fields here. Are there pros and cons associated with declaring them or not?

Part of me likes the explicitness of declaring them.

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -0,0 +1,91 @@
...
+                "forum_commented": 0,
+                "forum_responded": 0,
+                "problem_attempted": 43,
+                "problem_completed": 3,
+                "video_played": 6,
+                "forum_created": 0
+            },
+        }
+        from pprint import pprint
+        pprint(response.data)
```
mulby: don't forget to remove this!

mulby:
Finished my pass, let me know when the URLs are re-configured and I can take another look.

analytics_data_api/v0/views/learners.py
```diff
@@ -0,0 +1,88 @@
...
+    course_id = None
+
+    def get(self, request, *args, **kwargs):
+        self.course_id = request.QUERY_PARAMS.get('course_id', None)
+        if not self.course_id:
+            raise ParseError('Course ID was not specified')
+        try:
+            CourseKey.from_string(self.course_id)
+        except InvalidKeyError:
+            raise ParseError('Course ID malformed')
```
dan-f: I'm curious as to what code path is triggered by this `ParseError`. Does it result in an HTTP 400? If so, what's the message?

analytics_data_api/v0/tests/views/test_learners.py
dan-f: A recap from our huddle. Event names should be as follows:

- problem_attempts
- problems_attempted
- problems_completed
- videos_viewed
- discussions_contributed

analyticsdataserver/settings/base.py
```diff
@@ -271,7 +271,11 @@
 
 ENABLE_ADMIN_SITE = False
 
+# base url to generate link to learner profile page
```
dan-f: This should actually be the URL to generate a link to the User API, not the HTML page.

analytics_data_api/v0/tests/views/test_learners.py
dan-f: Great!

analytics_data_api/v0/models.py
dan-f: :+1: to this. Could we alternatively add it to the django settings module?

dan-f:
Let me know when you get to error messages stuff. Looking good!

analytics_data_api/v0/models.py
dsjen: That's interesting.  I would have expected the course ID to be wrapped in a `filter` and the username to nested in a `match` field.  As I understand, the filter component is a caching optimization.  I'm going to punt on that for now and just set these to be both `query` and `term` and the result should be the same.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: Thanks for the clarification.  I wasn't sure if by "invalid" you meant malformed (e.g. course ID of "not-a/course").  For malformed, we throw a 400.

analytics_data_api/v0/models.py
dsjen: Makes sense.  I'll update the comment.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: Oops!  I'll remove that.

analytics_data_api/v0/views/learners.py
dsjen: Yes, this returns a 400, but not in the edx recommended format.  I'll see about updating that.

analyticsdataserver/settings/base.py
dsjen: Ah, I see!  I'll update the comment.

analytics_data_api/v0/models.py
dsjen: Thanks, all!  I went with the `AppConfig.ready` method--seems to work well.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: Thank you!  Updated.

dsjen:
@dan-f & @mulby -- I believe I've addressed all comments now.  Of note is that I've added custom exceptions and middlewares to catch them, which return error codes and message.

analytics_data_api/v0/apps.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ApiAppConfig(AppConfig):
+
+    name = 'analytics_data_api.v0'
+
+    def ready(self):
+        super(ApiAppConfig, self).ready()
+        if settings.ELASTICSEARCH_LEARNERS_HOST:
+            connections.connections.create_connection(hosts=[settings.ELASTICSEARCH_LEARNERS_HOST])
```
dan-f: [This page](https://docs.djangoproject.com/en/1.9/ref/applications/#django.apps.AppConfig.ready) warns against interacting with the database in the `ready` method. Apparently it's called every time a management command is invoked. Since we're not querying Elasticsearch here, it's probably not an issue, right? cc @mulby

analytics_data_api/v0/exceptions.py
```diff
@@ -0,0 +1,52 @@
+import abc
+
+
+class BaseError(Exception):
+    """
+    Base error.
+    """
+
+    __metaclass__ = abc.ABCMeta
```
dan-f: Why do we need `abc`? I thought that the `Exception` class provides this behavior on its own.

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -0,0 +1,88 @@
...
+        user_name = 'a_user'
+        course_id = 'edX/DemoX/Demo_Course'
+        response = self.authenticated_get(self.path_template.format(user_name, course_id))
+        self.assertEquals(response.status_code, 404)
+
+    def test_get_400(self):
+        base_path = '/api/v0/learners/{}'
+        path = (base_path).format('ed_xavier')
+        response = self.authenticated_get(path)
+        self.assertEquals(response.status_code, 400)
```
dan-f: One final request: I'd prefer if the test cases are split up by edge cases rather than error codes. It makes the test cases more targeted and descriptive; we could have a `test_no_course_id`, and a `test_bad_course_id` rather than `test_get_400`, for example.

Also, now that we have awesome error handling, could you verify that the JSON corresponding to the error codes is correct?

dan-f:
Looks great to me, @dsjen. Thanks for being responsive to feedback! I love the use of middleware for error handling. :+1: once you address my two questions and one suggestion.

analytics_data_api/v0/apps.py
dsjen: Good point!  We're mocking out the calls in our unit tests, so this shouldn't be hit.

analytics_data_api/v0/exceptions.py
dsjen: I'm preventing others from creating an instance of `BaseError` directly and use the other concrete classes defined in the file.  I believe without `abc` once could still create an instance of an `BaseError` otherwise.

dan-f: I said it poorly, but I think what I meant was that the abc/message/__str__ stuff may be unnecessary because the base Exception class provides that functionality itself. I don't mind leaving this, though.

dsjen:
@dan-f -- thanks!  I've separated the tests out into two methods that verify the json payload against expected results.

PR #91, opened 2015-12-04
Run unit tests against real elasticsearch

dan-f:
Also removes the 'problem_attempts' field from the roster entry, since
there's no defined use case.

AN-6151

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,6 @@ script:
     - make validate
     - git fetch origin master:refs/remotes/origin/master # https://github.com/edx/diff-cover#troubleshooting
     - make diff.report
-branches:
-    only:
-      - master
 after_success:
     - coveralls
     - bash ./scripts/build-stats-to-datadog.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,15 @@ language: python
 python: "2.7"
 install:
     - "pip install coveralls"
+    - "make test.install_elasticsearch"
 cache: pip
 # Use docker for builds
 sudo: false
+before_script:
+    - make test.run_elasticsearch
+    # Recommended by Travis in order to make sure ElasticSearch
+    # actually starts up.  See # https://docs.travis-ci.com/user/database-setup/#ElasticSearch
+    - sleep 10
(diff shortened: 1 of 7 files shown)


dan-f:
@dsjen please review

dsjen:
@dan-f this is fantastic!  I run into a an `IndexAlreadyExistsException` when running locally though.  Are you seeing that with `make test`?

dan-f:
@dsjen that shouldn't happen... the test is supposed to delete the index at the end of the test. Are you positive that you're testing against test settings? e.g. `ELASTICSEARCH_LEARNERS_HOST` should be `'http://localhost:9200/'`.

dsjen:
I see.  I was running against the test settings, but I hadn't run `make test.install_elasticsearch` and `make test.run_elasticsearch`.  After doing that, tests pass.

Could we add elastic search start up and shut down when running `make test`?  It seems like it would also be advantageous to add `test.install_elasticsearch` to the `test.requirements` step too, but I'm not sure how easy it is to detect if elasticsearch has been downloaded.

dan-f:
I see. I think I can write `test.instsall_elasticsearch` such that it only installs if it hasn't already been installed on the system.

dan-f:
It turns out that testing for elasticsearch and automating it's start up are a little trickier than I initially thought. To test for elasticsearch or run it (as a system service), you need sudo privileges, which travis doesn't give us. Automating elasticsearch's startup is difficult because it takes a nondeterministic amount of time. I hacked around that with Travis' suggestion of putting a 'sleep 10' in the .travis.yml, but I don't like the idea of adding that to the makefile. How about I document in the README that ES must be running locally for tests to run? Once you start the ES server as a daemon, you can forget about it, rather than start and restart every time tests run. This is the approach that the [edx-notes-api](https://github.com/edx/edx-notes-api) takes.

dsjen:
@dan-f -- thanks for investigating that!  Adding documentation to the README sounds good to me.

dan-f:
Updated @dsjen!

dsjen:
Beautiful! :+1:

PR #92, opened 2015-12-08
Added learner engagement timeline.

dsjen:
@dan-f, this adds in the learn engagement timeline endpoint.  Please review when you get the chance.

### DIFF
--- a/analytics_data_api/v0/models.py
+++ b/analytics_data_api/v0/models.py
@@ -1,8 +1,11 @@
+from itertools import groupby
+
 from django.conf import settings
 from django.db import models
+from django.db.models import Sum
 from elasticsearch_dsl import DocType, Q
 
-from analytics_data_api.constants import country, genders
+from analytics_data_api.constants import country, engagement_entity_types, genders
 
 
 class CourseActivityWeekly(models.Model):
@@ -272,3 +275,54 @@ class RosterEntry(DocType):
         search = search.sort(sort_term)
 
         return search
+
+
+class ModuleEngagementTimelineManager(models.Manager):
+    """
+    Modifies the ModuleEngagement queryset to aggregate engagement data for
+    the learner engagement timeline.
+    """
+    def get_timelines(self, course_id, username):
+        queryset = ModuleEngagement.objects.all().filter(course_id=course_id, username=username) \
+            .values('date', 'entity_type', 'event') \
+            .annotate(count=Sum('count')) \
+            .order_by('date')
+
+        timelines = []
+
+        for key, group in groupby(queryset, lambda x: (x['date'])):
+            # Iterate over groups and create a single item with engagement data
+            item = {
+                u'date': key,
+            }
+            for engagement in group:
+                entity_type = engagement_entity_types.SINGULAR_TO_PLURAL[engagement['entity_type']]
+                engagement_type = '{}_{}'.format(entity_type, engagement['event'])
+                count = item.get(engagement_type, 0)
+                count += engagement['count']
+                item[engagement_type] = count
+            timelines.append(item)
+
+        return timelines
+
+
+class ModuleEngagement(models.Model):
+    """User interactions with entities within the courseware."""
+
+    course_id = models.CharField(db_index=True, max_length=255)
+    username = models.CharField(max_length=255)
+    date = models.DateTimeField()
+    # This will be one of "problem", "video" or "forum"
+    entity_type = models.CharField(max_length=255)
+    # For problems this will be the usage key, for videos it will be the html encoded module ID,
+    # for forums it will be the commentable_id
+    entity_id = models.CharField(max_length=255)
+    # A description of what interaction occurred, e.g. "contributed" or "viewed"
+    event = models.CharField(max_length=255)
+    # The number of times the user interacted with this entity in this way on this day.
+    count = models.IntegerField()
+
+    objects = ModuleEngagementTimelineManager()
+
+    class Meta(object):
+        db_table = 'module_engagement'
--- a/analytics_data_api/v0/tests/views/__init__.py
+++ b/analytics_data_api/v0/tests/views/__init__.py
@@ -1,4 +1,7 @@
+import json
+
 from opaque_keys.edx.keys import CourseKey
+from rest_framework import status
 
 DEMO_COURSE_ID = u'course-v1:edX+DemoX+Demo_2014'
 
@@ -12,3 +15,24 @@ class DemoCourseMixin(object):
         cls.course_id = DEMO_COURSE_ID
         cls.course_key = CourseKey.from_string(cls.course_id)
         super(DemoCourseMixin, cls).setUpClass()
+
+
+class VerifyCourseIdMixin(object):
+
+    def verify_no_course_id(self, response):
+        """ Assert that a course ID must be provided. """
+        self.assertEquals(response.status_code, status.HTTP_400_BAD_REQUEST)
+        expected = {
+            u"error_code": u"course_not_specified",
+            u"developer_message": u"Course id/key not specified."
+        }
+        self.assertDictEqual(json.loads(response.content), expected)
+
+    def verify_bad_course_id(self, response, course_id='malformed-course-id'):
+        """ Assert that a course ID must be valid. """
+        self.assertEquals(response.status_code, status.HTTP_400_BAD_REQUEST)
+        expected = {
+            u"error_code": u"course_key_malformed",
+            u"developer_message": u"Course id/key {} malformed.".format(course_id)
+        }
+        self.assertDictEqual(json.loads(response.content), expected)
--- a/analytics_data_api/v0/urls/__init__.py
+++ b/analytics_data_api/v0/urls/__init__.py
@@ -2,12 +2,16 @@ from django.conf.urls import patterns, url, include
 from django.core.urlresolvers import reverse_lazy
 from django.views.generic import RedirectView
 
+USERNAME_PATTERN = r'(?P<username>.+)'
+
 urlpatterns = patterns(
     '',
     url(r'^courses/', include('analytics_data_api.v0.urls.courses', namespace='courses')),
     url(r'^problems/', include('analytics_data_api.v0.urls.problems', namespace='problems')),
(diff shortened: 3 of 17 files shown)


analytics_data_api/v0/urls/engagement_timelines.py
```diff
@@ -0,0 +1,14 @@
...
+
+ENGAGEMENT_TIMELINE_URLS = [
+    ('', views.EngagementTimelineView, 'engagement_timelines')
+]
+
+urlpatterns = []
+
+for path, view, name in ENGAGEMENT_TIMELINE_URLS:
+    regex = r'^{0}/$'.format(USERNAME_PATTERN)
+    urlpatterns += patterns('', url(regex, view.as_view(), name=name))
```
dan-f: I'm not a huge fan of this pattern for constructing URLs - I find that at a quick glance it's hard to tell what the URLs in this file are, especially since there's just a single url, which is r'^(?P<username>.+)/$'. The pattern I'm used to is as follows (the example is `urls/learners.py` taken from my learner list PR):

```python
from django.conf.urls import patterns, url

from analytics_data_api.v0.views import learners as views


USERNAME_PATTERN = r'(?P<username>.+)'

urlpatterns = patterns(
    '',
    url(r'^$', views.LearnerListView.as_view(), name='learners'),
    url(r'^{}/$'.format(USERNAME_PATTERN), views.LearnerView.as_view(), name='learner'),
)
```

analytics_data_api/v0/urls/__init__.py
```diff
@@ -2,12 +2,16 @@
 from django.core.urlresolvers import reverse_lazy
 from django.views.generic import RedirectView
 
+USERNAME_PATTERN = r'(?P<username>.+)'
```
dan-f: nice!

analytics_data_api/v0/models.py
```diff
@@ -220,3 +223,64 @@ class Meta:
...
+class ModuleEngagement(models.Model):
+    """User interactions with entities within the courseware."""
+
+    course_id = models.CharField(db_index=True, max_length=255)
+    username = models.CharField(max_length=255)
+    date = models.DateTimeField()
+    # This will be one of "problem", "video" or "forum"
+    entity_type = models.CharField(max_length=255)
+    # For problems this will be the usage key, for videos it will be the html encoded module ID,
+    # for forums it will be the commentable_id
```
dan-f: Great comments. Thanks!

analytics_data_api/v0/models.py
```diff
@@ -220,3 +223,64 @@ class Meta:
...
+    entity_id = models.CharField(max_length=255)
+    # A description of what interaction occurred.
```
dan-f: Nit: perhaps an example?

analytics_data_api/v0/models.py
```diff
@@ -220,3 +223,64 @@ class Meta:
...
+        db_table = 'module_engagement'
+
+
+class ModuleEngagementTimelineManager(models.Manager):
+    """
+    Modifies the ModuleEngagement queryset to aggregate engagement data for
+    the learner engagement timeline.
+    """
+    def get_timelines(self, course_id, username):
+        queryset = ModuleEngagement.objects.all().filter(course_id=course_id, username=username)\
```
dan-f: obsessive nit: space before `\` on these lines

dan-f:
Two things regarding the model manager for the timeline:

* Can we provide the `get_timelines` method with start and end dates? That way we limit the size of the result set returned by the query.
* Can we use ORM methods to get the timeline data in the format we want, so that we don't have to do an O(N) scan in python of the timeline results?

analytics_data_api/v0/models.py
```diff
@@ -220,3 +223,64 @@ class Meta:
...
+class ModuleEngagementTimeline(models.Model):
+    """
+    Learner engagement timeline data.
+    """
+    date = models.DateField()
+    problems_attempted = models.IntegerField()
+    problems_completed = models.IntegerField()
+    discussions_contributed = models.IntegerField()
+    videos_viewed = models.IntegerField()
+    objects = ModuleEngagementTimelineManager()
```
dan-f: I'm used to the django convention of the manager belonging to the model it manages. If the `ModulelEngagementTimelineManager` were to belong to the `ModuleEngagement` class, would we be able to get rid of this model?

I'm also skeptical of having a model that doesn't map to a database table.

dan-f:
Is `settings.DATE_FORMAT` ISO 8601?(https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions#edXRESTAPIConventions-12.SerializationConventions)

analytics_data_api/v0/models.py
```diff
@@ -220,3 +223,64 @@ class Meta:
...
+    """
+    def get_timelines(self, course_id, username):
+        queryset = ModuleEngagement.objects.all().filter(course_id=course_id, username=username)\
+            .values('date', 'entity_type', 'event')\
+            .annotate(count=Sum('count'))\
+            .order_by('date')
+
+        timelines = []
+
+        for key, group in groupby(queryset, lambda x: (x['date'])):
```
dan-f: Performance conern:

We have (up to) `n` days and `m` modules that a student interacts with on any given day. Theoretically, the queryset is as large as `O(n * m)`, and I assume that `itertools.groupby` could be an `O(n^2)` operation, so we may have `O((n * m)^2` here.

@mulby and I chatted about this, and he's optimistic that in practice, `m` will be very small, so he's not terribly worried. Regardless, we explored a few ways to improve the performance of this code:

* Push this group by logic into SQL. We might be able to do this by making the above query a sub-query to a SELECT ... GROUP_BY clause. We're not sure how the `entity_type`/`event` joining logic would work, though.
* If we can't push the logic into SQL, find out if we can stream results out of SQL (do querysets work that way?) into memory, and manage a `dict` which maps date strings to engagement counts, thereby reducing the space complexity. Instead of the quadratic group_by, you can just iterate through the results one at a time, either adding a new date, or updating an existing one. For example, the dict would have the form:
```python
{
    '01-01-01': {
        'problems_attempted': 0,
        ....
    },
    ...
}
```

dan-f:
@dsjen I updated the API doc, removing the `date_range` field from the timeline, since it's redundant. @mulby and I thought that `date_range` still has value in the endpoints which return aggregate engagement data for context.

analytics_data_api/v0/tests/views/__init__.py
```diff
@@ -12,3 +15,27 @@ def setUpClass(cls):
         cls.course_id = DEMO_COURSE_ID
         cls.course_key = CourseKey.from_string(cls.course_id)
         super(DemoCourseMixin, cls).setUpClass()
+
+
+class VerifyNoCourseIdMixin(object):
```
dan-f: Why not add both these `verify` methods to the same mixin? Perhaps the `VerifyCourseIdMixin`?

In general, I think that if a mixin has only a single method, that method might as well be a function not bound to an object or class.

dan-f: Alternatively, something like [`test_bad_request`](https://github.com/edx/edx-analytics-data-api/pull/93/files#diff-7bb85fd0aed6837c1d0d4834f7ff1878R412) might be a little more generic/reusable.

dan-f:
A few final comments:

* Were we going to support configurable start/end dates for the timeline in V0?
* Can you make sure the following test cases are covered?
  * No username provided
  * Engagements default to 0 for a given day when there are no rows for that particular day/engagement in the database.
  * What if there's no engagement data for a particular day? Should we simply not include that day in the array? Should we include that day with 0s?
  * What if there's no data at all? Is this the 404 case? I'm wondering about the pros/cons of 404 vs an empty list.

analytics_data_api/v0/urls/engagement_timelines.py
dsjen: That makes sense.  Updated!

analytics_data_api/v0/models.py
dsjen: Thanks, updated.

analytics_data_api/v0/models.py
dsjen: :) Consider it done.

analytics_data_api/v0/models.py
dsjen: Yes!  Moving `objects` into `ModuleEngagement` rather.  That feels much better now.

dsjen:
@dan-f  Thanks!  Yes, `settings.DATE_FORMAT` ISO 8601--e.g. YYYY-MM-DD.

I'm not sure that we need to test explicitly for the no username provided case because that isn't a valid route as defined (or by not defining it) in our urls.py file.

We currently return a 404 when no data is available along with a error code and developer message.  I'm not opposed to an empty list either, but the error code does make it clear why a 404 is returned.

As discussed offline, I'll in some comments for that days will be skipped when there is no data for those days.

Please let me know what you think!  I'm moving forward with updating the payload according to the revised spec, updating comments, and combining the test mixins (I had grand plans for the single method mixins that fizzled out!).

I'll let you know when I've updated the PR.

dsjen:
@dan-f -- I've updated the PR (mostly with deletions :smile:).  Please let me know if there's anything else.

analytics_data_api/v0/views/engagement_timelines.py
```diff
@@ -0,0 +1,56 @@
...
+
+    """
+    serializer_class = EngagementDaySerializer
+    username = None
+    lookup_field = 'username'
+
+    def list(self, request, *args, **kwargs):
+        response = super(EngagementTimelineView, self).list(request, *args, **kwargs)
+        if response.status_code == status.HTTP_200_OK:
+            response.data = {'days': response.data}
```
dan-f: There it is! Sneaky!

dan-f:
Could you make sure to document and test the case in which there's no data for 1 or more days?

dsjen:
Updated to include a test for a gap day and documentation added to the endpoint.  I'm going to rebase (wish me luck).

dsjen:
@dan-f -- rebased.  Please let me know if there's anything else.

dan-f:
thanks @dsjen! LGTM!

PR #94, opened 2015-12-11
Added learner course metadata

dsjen:
@dan-f -- here's a PR for our work on the learner course metadata endpoint.  There is a pylint error that I'm looking into fixes.  I'll update when that is fixed.


dsjen:
@dan-f -- build is passing!  I'm starting my review now.

analytics_data_api/v0/models.py
```diff
@@ -242,8 +242,22 @@ def get_users_in_course(
         the Search object.  Raises `ValueError` if both `segments` and
         `ignore_segments` are provided.
         """
+        # Error handling
         if segments and ignore_segments:
             raise ValueError('Cannot combine `segments` and `ignore_segments` parameters.')
+        for segment in (segments or list()) + (ignore_segments or list()):
+            if segment not in learner.SEGMENTS:
+                raise ValueError(
+                    'segments/ignore_segments values must belong to: {}'.format(', '.join(learner.SEGMENTS))
```
dsjen: It might be helpful to state the invalid segment.  Maybe something like: `segments/ignore_segments value ('{segment}') must belong to: {the_list}`

analytics_data_api/v0/models.py
```diff
@@ -242,8 +242,22 @@ def get_users_in_course(
...
+                )
+        order_by_options = (
+            'username', 'email', 'discussions_contributed', 'problems_attempted', 'problems_completed', 'videos_viewed'
+        )
+        sort_order_options = ('asc', 'desc')
+        if order_by not in order_by_options:
+            raise ValueError('order_by value must be one of: {}'.format(', '.join(order_by_options)))
```
dsjen: Same suggestion to list the invalid argument.

analytics_data_api/v0/models.py
```diff
@@ -242,8 +242,22 @@ def get_users_in_course(
...
+        if sort_order not in sort_order_options:
+            raise ValueError('sort_order value must be one of: {}'.format(', '.join(sort_order_options)))
```
dsjen: Same suggestion to list the invalid argument.

analytics_data_api/v0/serializers.py
```diff
@@ -388,3 +388,83 @@ def transform_discussions_contributed(self, _obj, value):
...
+
+
+class DateRangeSerializer(serializers.Serializer):
+    start = serializers.DateTimeField(source='start_date', format=settings.DATE_FORMAT)
+    end = serializers.DateTimeField(source='end_date', format=settings.DATE_FORMAT)
+
+
+class EnagementRangeMetricSerializer(serializers.Serializer):
+    """
+    TODO: TBD
```
dsjen: Note for Dennis: Document this.

analytics_data_api/v0/models.py
```diff
@@ -263,19 +277,54 @@ def get_users_in_course(
...
+                <segment_name>: <learner_count>
+            },
+            'enrollment_modes': {
+                <enrollment_mode_name>: <learner_count>
+            }
+        }
+        """
+        search = cls.search()
+        search.query = Q('bool', must=[Q('term', course_id=course_id)])
+        search.aggs.bucket('enrollment_modes', 'terms', field='enrollment_mode')
```
dsjen: This is up to you, but it might be a bit more succinct to wrap this in a loop: ```search.aggs.bucket(mode, 'terms', field=mode) for mode in ['enrollment_modes', 'segments', 'cohorts']``` (inspired by inline loops below).

analytics_data_api/v0/models.py
```diff
@@ -263,19 +277,54 @@ def get_users_in_course(
...
+                bucket.key: bucket.doc_count
+                for bucket in response.aggregations[aggregation_name].buckets
+            }
+            for aggregation_name in response.aggregations
+        }
+        # Add default values of 0 for segments with no learners
+        default_segments = {
+            segment: 0 for segment in learner.SEGMENTS
+        }
+        for segment in default_segments:
```
dsjen: Is `default_segments` needed?  It seems like the loop could iterate over `learner.SEGMENTS` and then default to `0`.

analytics_data_api/v0/views/learners.py
```diff
@@ -179,3 +191,99 @@ def get_queryset(self):
...
+
+    **Example Request**
+
+        GET /api/v0/course_learner_metadata/{course_id}/
+
+    **Response Values**
+
+        Returns a JSON object with the following keys:
+
+            * cohorts: A JSON object mapping the names of cohorts in the course
```
dsjen: Nit: it doesn't seem necessary to restate that each key is a JSON object.

analytics_data_api/v0/views/learners.py
```diff
@@ -179,3 +191,99 @@ def get_queryset(self):
...
+              to the number of students belonging to those cohorts.
+            * segments: A JSON object mapping the names of segments in the
```
dsjen: Example segments might be helpful.

analytics_data_api/v0/views/learners.py
```diff
@@ -179,3 +191,99 @@ def get_queryset(self):
...
+              course to the number of students belonging to those segments.
+            * enrollment_modes: A JSON object mapping the names of enrollment
```
dsjen: Example enrollment modes might be helpful.

analytics_data_api/v0/views/learners.py
```diff
@@ -179,3 +191,99 @@ def get_queryset(self):
...
+              the following keys:
+                * date_range: The time duration for which this data applies
+                * problems_attempted: engagement ranges for the number of
+                  problems attempted in the date range.
+                * problems_completed: engagement ranges for the number of
+                  problems completed in the date range.
+                * problems_attempted_per_completed: engagement ranges for the
+                  number of problems attempted per completed in the date range.
+                * discussions_contributed: engagement ranges for the number of
+                  discussions contributed per completed in the date range.
```
dsjen: extra `per completed`

dsjen:
@dan-f -- I don't see a test for when there's data for when all fields (e.g. cohort, segment, enrollment_mode) are populated.

analytics_data_api/v0/models.py
```diff
@@ -326,3 +375,21 @@ class ModuleEngagement(models.Model):
 
     class Meta(object):
         db_table = 'module_engagement'
+
+
+class ModuleEngagementMetricRanges(models.Model):
+    """Engagement metric ranges."""
```
dan-f: This docstring could be a little more descriptive.

analytics_data_api/v0/serializers.py
```diff
@@ -388,3 +388,73 @@ def transform_discussions_contributed(self, _obj, value):
...
+    Serializes ModuleEngagementMetricRanges (low_range and high_range) into
+    the below_average, average, above_average ranges represented as arrays.
+    """
+    below_average = serializers.SerializerMethodField('get_below_average_range')
+    average = serializers.SerializerMethodField('get_average_range')
+    above_average = serializers.SerializerMethodField('get_above_average_range')
+
+    def get_average_range(self, obj):
+        metric_range = [
+            obj['low_range'].high_value if obj['low_range'] else None,
```
dan-f: Can we always assume that the object has both `low_value` and `high_value` attributes? what about the case of an unbounded (i.e. infinite) value?

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -404,9 +412,192 @@ def test_pagination(self):
...
     )
     @ddt.unpack
     def test_bad_request(self, parameters, expected_error_code):
         response = self.authenticated_get('/api/v0/learners/', parameters)
         self.assertEqual(response.status_code, 400)
         self.assertEqual(json.loads(response.content)['error_code'], expected_error_code)
+
+
+@ddt.ddt
+class CourseLearnerMetadataTests(DemoCourseMixin, VerifyCourseIdMixin,
```
dan-f: I had to remind myself why we're calling the mixins first in the MRO (usually they're called after the parent class), which I thought was because of this bug - https://code.djangoproject.com/ticket/15828, but it looks like it was fixed a while ago. Does this work when the mixins come after the `TestCaseWithAuthentication`?

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -404,9 +412,192 @@ def test_pagination(self):
...
+        for entity_type in engagement_entity_types.AGGREGATE_TYPES:
+            for event in engagement_events.EVENTS[entity_type]:
+                metrics.append('{0}_{1}'.format(entity_type, event))
+        return metrics
+
+    # TODO: enable during https://openedx.atlassian.net/browse/AN-6319
+    # def test_cohorts(self):
+    #     pass
+
+    def test_no_engagement_ranges(self):
```
dan-f: I'm curious about this. The way we handled the 'no data' case with the timeline view was to return a 404 here. Then again, I think this makes sense, because we want to still be able to get segments and other data from this endpoint even if not all the data is available.

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -404,9 +412,192 @@ def test_pagination(self):
...
+        response = self._get(self.course_id)
+        self.assertEqual(response.status_code, 200)
+        self.assertDictContainsSubset(self.empty_engagement_ranges, json.loads(response.content))
+
+    def test_one_engagement_range(self):
+        metric_type = '{0}_{1}'.format(engagement_entity_types.PROBLEMS, engagement_events.COMPLETED)
```
dan-f: Nit: I'd prefer just "problems_completed" here, since it's 1) more readable and 2) will mean that tests will fail if someone accidentally changes the value of `engagement_entity_types` or `engagement_events `.

analytics_data_api/v0/serializers.py
dsjen: I checked on this and the values will either be NULL or a number--no infinite values.

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -404,9 +412,192 @@ def test_pagination(self):
...
+                'date_range': {
+                    'start': '2015-07-01',
+                    'end': '2015-07-21'
+                }
+            }
+        }
+
+        max_value = 1000.0
+        for metric_type in self.engagement_metrics:
+            low_ceil = random.random() * max_value * 0.5
```
dan-f: I'd prefer if we didn't use the `random` module for values in tests, because I don't like the idea of nondeterministic tests. If you're concerned about a variety of values, then you can parameterize this method and use `ddt` in the test cases for particularly problematic values, but if there's no reason for variance in these values, then I don't see a problem with hardcoding them in.

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -404,9 +412,192 @@ def test_pagination(self):
...
+
+    def test_segments_same_learner(self):
+        """
+        Tests segment counts when each learner belongs to multiple segments.
+        """
+        self.create_learners([
+            {'username': 'user_1', 'course_id': self.course_id, 'segments': ['struggling', 'disengaging']},
+            {'username': 'user_2', 'course_id': self.course_id, 'segments': ['disengaging']}
+        ])
+        expected = self._get_full_engagement_ranges()
```
dan-f: I think this could use some cleanup since we both touched these test cases. I could do it or you could, but I'll hold off for now in case `self._get_full_engagement_ranges` changes.

analytics_data_api/v0/urls/learners.py
```diff
@@ -1,11 +1,17 @@
 from django.conf.urls import patterns, url
 
+from analytics_data_api.v0.urls import COURSE_ID_PATTERN
 from analytics_data_api.v0.views import learners as views
-from analytics_data_api.v0.urls import USERNAME_PATTERN
+
+USERNAME_PATTERN = r'(?P<username>.+)'
```
dan-f: I just noticed this, but is our `USERNAME_PATTERN` really `.+`? Seems pretty indiscriminate to me. [Here's the one we use in edx-platform](https://github.com/edx/edx-platform/blob/69c47e09511c0fda4762dc959a6a3025dc96f272/openedx/core/djangoapps/user_api/urls.py#L11). We don't have to address this in the current PR, but I wanted to make note regardless.

analytics_data_api/v0/models.py
dsjen: Updated. :)

analytics_data_api/v0/views/__init__.py
```diff
@@ -6,13 +6,16 @@
 
 class CourseViewMixin(object):
     """
-    Captures the course_id query arg and validates it.
+    Captures the course_id from arguments or from the query arg and validates it.
```
dan-f: From the URL?

analytics_data_api/v0/views/__init__.py
```diff
@@ -6,13 +6,16 @@
...
     """
 
     course_id = None
 
     def get(self, request, *args, **kwargs):
-        self.course_id = request.QUERY_PARAMS.get('course_id', None)
+        self.course_id = self.kwargs.get('course_id', None)
```
dan-f: These lines could be squashed into:

```python
self.course_id = self.kwargs.get('course_id', request.QUERY_PARAMS.get('course_id', None))
```

dan-f:
@dsjen finished my review of your part. A few comments, but overall things look good! I'll start addressing your review.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: I tested this out and moved the base class after the mixins to no avail--a bunch of errors indicating setUp wasn't called appeared.

analytics_data_api/v0/models.py
dan-f: It'd take a bit more work to get the bucket keys I want, and since it'd only save one line, I'm going to leave it as is.

analytics_data_api/v0/views/__init__.py
dsjen: Oh, I like that!

analytics_data_api/v0/models.py
dan-f: Yeah, good point!

analytics_data_api/v0/views/__init__.py
dsjen: Updated!

analytics_data_api/v0/views/learners.py
dan-f: I changed it to "An object".

analytics_data_api/v0/urls/learners.py
dsjen: Thanks for the linked.  I've updated the regex.

analytics_data_api/v0/views/learners.py
dan-f: Good call!

analytics_data_api/v0/tests/views/test_learners.py
dsjen: I was feeling conflicted about that too and will update to not use random values.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: That's a good point!  Updated.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: Agreed!

analytics_data_api/v0/models.py
dsjen: Sounds good.

analytics_data_api/v0/tests/views/test_learners.py
dan-f: blah! Oh well thanks for testing!

dan-f:
@dsjen thanks for addressing my feedback. I also pushed up a commit addressing yours yesterday - let me know if you're OK with it. I think I'd like to take a crack at refactoring the test methods a little bit, but after that this gets a :+1: from me!

dsjen:
@dan-f -- thanks for the review and updates!  LGTM.  :+1:

Please let me know if you'd like help with the test methods refactor.

dan-f:
Will do. I will at least need some help when it comes to squashing these commits down. I'll let you know

PR #99, opened 2015-12-23
Added last_updated field to learner endpoints.

dsjen:
This introduces a new index that stories the last updated date for the learn roster entries index.

@dan-f, please review.


analytics_data_api/v0/serializers.py
```diff
@@ -361,19 +362,6 @@ class EdxPaginationSerializer(pagination.PaginationSerializer):
     num_pages = serializers.Field(source='paginator.num_pages')
 
 
-class ElasticsearchDSLSearchSerializer(EdxPaginationSerializer):
```
dan-f: Why did you remove this?

analytics_data_api/v0/views/learners.py
```diff
@@ -223,7 +233,13 @@ def get_queryset(self):
         # parameter values in `get_users_in_course`.
         params = {key: val for key, val in params.items() if val is not None}
         try:
-            return RosterEntry.get_users_in_course(self.course_id, **params)
+            users = RosterEntry.get_users_in_course(self.course_id, **params)
+            last_updated = {'date': None}
+            if len(RosterUpdate.get_last_updated()) == 1:
+                last_updated = RosterUpdate.get_last_updated()[0]
```
dan-f: Could you refactor this so that the `RosterUpdate.get_last_updated` query is only run once?

analytics_data_api/v0/views/learners.py
```diff
@@ -223,7 +233,13 @@ def get_queryset(self):
...
+            else:
+                logger.warn('RosterUpdate not found.')
```
dan-f: Is it possible for there to be no data for the requested course in the `ELASTICSEARCH_LEARNERS_UPDATE_INDEX`, but data in the `ELASTICSEARCH_LEARNERS_INDEX`?

analytics_data_api/v0/serializers.py
```diff
@@ -426,20 +414,11 @@ def _get_range(self, metric_range):
 
 
 class CourseLearnerMetadataSerializer(serializers.Serializer):
-    enrollment_modes = serializers.SerializerMethodField('get_enrollment_modes')
-    segments = serializers.SerializerMethodField('get_segments')
-    cohorts = serializers.SerializerMethodField('get_cohorts')
+    enrollment_modes = serializers.Field(source='es_data.enrollment_modes')
```
dan-f: :100: !!!!

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -87,6 +102,7 @@ def tearDown(self):
         """Remove the index after every test."""
         super(LearnerAPITestMixin, self).tearDown()
         self._es.indices.delete(index=settings.ELASTICSEARCH_LEARNERS_INDEX)
+        self._es.indices.delete(index=settings.ELASTICSEARCH_LEARNERS_UPDATE_INDEX)
```
dan-f: I know it's not directly related to this PR, but would you mind refactoring this cleanup to use [`self.addCleanup`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.addCleanup)? I recently learned that it's preferred over `self.tearDown`, since every statement in tearDown may not run if one of them fails.

analytics_data_api/v0/views/learners.py
```diff
@@ -223,7 +233,13 @@ def get_queryset(self):
...
+            return [{'learner': user, 'last_updated': last_updated} for user in users.execute()]
```
dan-f: I think I'd prefer to see this logic handled at the model layer. What do you think?

dan-f:
I'm confused about the implementation here of the concept of "last updated". My understanding was that it would apply to the particular data that is being requested, but currently all endpoints will share the same last updated value. Is it true that "last updated" will be the same date across the board @mulby?

analytics_data_api/v0/views/learners.py
dan-f: Also, this implementation breaks the pagination values code path; for the pagination serializer to add `count`, `next`, `previous`, and `num_pages` it calls into the `elasticsearch_dsl` `Search` object. Now that we have a python list instead, `count` no longer is the number of matching results, `next` and `previous` are `null`, and `num_pages` is 1.

...why didn't the unit tests break?

mulby:
@dan-f all data in the index should be updated at roughly the same time, so we save ourselves the overhead of an additional field on every record by just recording the time in one location. The timing doesn't need to be exact so this should be "good enough".

This should only apply to the elasticsearch data. Hopefully that's the case here.

analytics_data_api/v0/serializers.py
dsjen: :hamster: I can't tell you how excited I was to find this too!

analytics_data_api/v0/tests/views/test_learners.py
dsjen: That's very interesting!  I'll update this.

analytics_data_api/v0/views/learners.py
dsjen: I don't mind updating this.  FYI, the call to `execute` is cached according to https://elasticsearch-dsl.readthedocs.org/en/latest/search_dsl.html.

analytics_data_api/v0/views/learners.py
dsjen: The intension is for for update index to complement the learn index, and my thinking here is that it doesn't make sense for the entire call to fail if the update index isn't available for some reason.  What do you think?

analytics_data_api/v0/views/learners.py
dsjen: Oh!  I tested this out quickly and saw that we were getting results for those fields (empty it turns out) and thought everything was working as expected.

I agree about moving the `execute()` call to the model--actually had intended on doing that.  I'll bring back `ElasticsearchDSLSearchSerializer` and see if I can figure out what went awry with the tests; oddly, they passed.

analytics_data_api/v0/views/learners.py
dan-f: Good to know.

analytics_data_api/v0/views/learners.py
dan-f: I agree, but I wanted to bring it up just in case.

dsjen:
@dan-f, I've pushed a commit to address your comments.  There didn't seem to be a way to make it all just work from within the serializer, so I ended up moving the `last_updated` field into its own serializer and updating the responses with the serialized data from the update index in the `list()` and `retrieve()` methods.

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -25,10 +25,16 @@ def setUp(self):
...
-        if self._es.indices.exists(index=settings.ELASTICSEARCH_LEARNERS_INDEX):
-            self._es.indices.delete(index=settings.ELASTICSEARCH_LEARNERS_INDEX)
-        self._es.indices.create(index=settings.ELASTICSEARCH_LEARNERS_INDEX)
+
+        for index in [settings.ELASTICSEARCH_LEARNERS_INDEX, settings.ELASTICSEARCH_LEARNERS_UPDATE_INDEX]:
+            # delete index if for some reason the index wasn't deleted
+            def delete_index(to_delete):
+                if self._es.indices.exists(index=to_delete):
+                    self._es.indices.delete(index=to_delete)
+            delete_index(index)
```
dan-f: Why do we need to make this call in setup if the indexes are always deleted during cleanup?

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -141,6 +152,20 @@ def create_learners(self, learners):
             self._create_learner(**learner)
         self._es.indices.refresh(index=settings.ELASTICSEARCH_LEARNERS_INDEX)
 
+    def create_index_update(self, date=None):
```
dan-f: Nit: rename to `create_update_index` since it's the update index?

analytics_data_api/v0/tests/views/test_learners.py
```diff
@@ -172,8 +197,8 @@ def test_get_user(self, username, name, course_id, enrollment_mode, segments=Non
             "problem_attempts_per_completed": problem_attempts_per_completed,
             "attempt_ratio_order": attempt_ratio_order,
             "enrollment_date": enrollment_date,
-            "last_updated": last_updated,
         }])
+        self.create_index_update(last_updated)
```
dan-f: Rather than call this in each test case, could you add it to `setUp`?

analytics_data_api/v0/views/learners.py
```diff
@@ -186,11 +215,20 @@ def _validate_query_params(self):
...
                     raise ParameterValueError('Page size must be in the range [1, {}]'.format(self.max_paginate_by))
 
+    def list(self, request, *args, **kwargs):
+        """
+        Adds the last_updated field to the results.
+        """
+        response = super(LearnerListView, self).list(request, args, kwargs)
+        last_updated = self.get_last_updated()
+        for result in response.data['results']:
+            result.update(last_updated)
```
dan-f: This differs from the current API spec, where `last_updated` is a key in the top-level JSON object. However, I don't really mind this implementation, and it makes things more consistent. Would you mind updating the API spec?

analytics_data_api/v0/tests/views/test_learners.py
dsjen: I was debating whether to include it or not.  I decided to include it in case the index wasn't deleted before our new snazzy `addCleanup` functionality.  I'm thinking now that we can assume that the system is clean and that there isn't a need for calling `delete_index()` explicitly; I'll remove it.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: Will do!

dan-f:
Looking good! :+1: once my final comments are addressed.

analytics_data_api/v0/tests/views/test_learners.py
dsjen: Right on!

analytics_data_api/v0/views/learners.py
dsjen: Thanks, updated!

PR #101, opened 2016-01-04
Renamed learner API field.

dsjen:
- discussions_contributed renamed to discussion_contributions

@dan-f please review


dan-f:
:+1: thanks!

PR #103, opened 2016-01-08
Added AWS request signing.

dsjen:
This PR adds a number of settings used for signing requests for when the elasticsearch cluster is hosted on aws.  It's still not entirely clear whether we want this in a separate repo for sharing yet, but I figure we can pull out `connections.py` into a package without too much trouble if we decide to go that route.

@dan-f, please review.

### DIFF
--- a/analytics_data_api/utils.py
+++ b/analytics_data_api/utils.py
@@ -1,9 +1,7 @@
-from collections import defaultdict
+from importlib import import_module
 
 from django.db.models import Q
 from rest_framework.authtoken.models import Token
-
-from analytics_data_api.v0.models import ProblemResponseAnswerDistribution
 
 
 def delete_user_auth_token(username):
@@ -47,49 +45,6 @@ def matching_tuple(answer):
     )
 
 
-def consolidate_answers(problem):
-    """ Attempt to consolidate erroneously randomized answers. """
-    answer_sets = defaultdict(list)
-    match_tuple_sets = defaultdict(set)
-
-    for answer in problem:
-        answer.consolidated_variant = False
-
-        answer_sets[answer.value_id].append(answer)
-        match_tuple_sets[answer.value_id].add(matching_tuple(answer))
-
-    # If a part has more than one unique tuple of matching fields, do not consolidate.
-    for _, match_tuple_set in match_tuple_sets.iteritems():
-        if len(match_tuple_set) > 1:
-            return problem
-
-    consolidated_answers = []
-
-    for _, answers in answer_sets.iteritems():
-        consolidated_answer = None
-
-        if len(answers) == 1:
-            consolidated_answers.append(answers[0])
-            continue
-
-        for answer in answers:
-            if consolidated_answer:
-                if isinstance(consolidated_answer, ProblemResponseAnswerDistribution):
-                    consolidated_answer.count += answer.count
-                else:
-                    consolidated_answer.first_response_count += answer.first_response_count
-                    consolidated_answer.last_response_count += answer.last_response_count
-            else:
-                consolidated_answer = answer
-
-                consolidated_answer.variant = None
-                consolidated_answer.consolidated_variant = True
-
-        consolidated_answers.append(consolidated_answer)
-
-    return consolidated_answers
-
-
 def dictfetchall(cursor):
     """Returns all rows from a cursor as a dict"""
 
@@ -98,3 +53,10 @@ def dictfetchall(cursor):
         dict(zip([col[0] for col in desc], row))
         for row in cursor.fetchall()
     ]
+
+
+def load_fully_qualified_definition(definition):
+    """ Returns the class given the full definition. """
+    module_name, class_name = definition.rsplit('.', 1)
+    module = import_module(module_name)
+    return getattr(module, class_name)
--- a/analytics_data_api/v0/connections.py
+++ b/analytics_data_api/v0/connections.py
@@ -1,0 +1,74 @@
+import json
+import time
+
+from boto.connection import AWSAuthConnection
+from elasticsearch import Connection
+
+
+class BotoHttpConnection(Connection):
+    """
+    Uses AWS configured connection to sign requests before they're sent to
+    elasticsearch nodes.
+    """
+
+    connection = None
+
+    def __init__(self, host='localhost', port=443, aws_access_key_id=None, aws_secret_access_key=None,
+                 region=None, **kwargs):
+        super(BotoHttpConnection, self).__init__(host=host, port=port, **kwargs)
+        connection_params = {'host': host, 'port': port}
+
+        # If not provided, boto will attempt to use default environment variables to fill
+        # the access credentials.
+        connection_params['aws_access_key_id'] = aws_access_key_id
+        connection_params['aws_secret_access_key'] = aws_secret_access_key
+        connection_params['region'] = region
+        # Remove 'None' values so that we don't overwrite defaults
+        connection_params = {key: val for key, val in connection_params.items() if val is not None}
+        self.connection = ESConnection(**connection_params)
+
+    # pylint: disable=unused-argument
+    def perform_request(self, method, url, params=None, body=None, timeout=None, ignore=()):
+        """
+        Called when making requests elasticsearch.  Requests are signed and
+        http status, headers, and response is returned.
+
+        Note: the "timeout" kwarg is ignored in this case.  Boto manages the timeout
+        and the default is 70 seconds.
+        See: https://github.com/boto/boto/blob/develop/boto/connection.py#L533
+        """
+        if not isinstance(body, basestring):
+            body = json.dumps(body)
+        start = time.time()
+        response = self.connection.make_request(method, url, params=params, data=body)
+        duration = time.time() - start
+        raw_data = response.read()
+
+        # raise errors based on http status codes and let the client handle them
+        if not (200 <= response.status < 300) and response.status not in ignore:
+            self.log_request_fail(method, url, body, duration, response.status)
+            self._raise_error(response.status, raw_data)
+
+        self.log_request_success(method, url, url, body, response.status, raw_data, duration)
+
+        return response.status, dict(response.getheaders()), raw_data
+
+
+class ESConnection(AWSAuthConnection):
+    """
+    Use to sign requests for an AWS hosted elasticsearch cluster.
+    """
+
+    def __init__(self, *args, **kwargs):
+        region = kwargs.pop('region', None)
+        kwargs.setdefault('is_secure', True)
+        super(ESConnection, self).__init__(*args, **kwargs)
+        self.auth_region_name = region
+        self.auth_service_name = 'es'
+
+    def _required_auth_capability(self):
+        """
+        Supplies the capabilities of the auth handler and signs the responses to
+        AWS using HMAC-4.
+        """
+        return ['hmac-v4']
--- a/analytics_data_api/v0/tests/test_connections.py
+++ b/analytics_data_api/v0/tests/test_connections.py
@@ -1,0 +1,66 @@
+import socket
+
+from django.test import TestCase
+from elasticsearch.exceptions import ElasticsearchException
+from mock import patch
+
+from analytics_data_api.v0.connections import BotoHttpConnection, ESConnection
+
+
+class ESConnectionTests(TestCase):
+
+    def test_constructor_params(self):
+        connection = ESConnection('mockservice.cc-zone-1.amazonaws.com',
+                                  aws_access_key_id='access_key',
+                                  aws_secret_access_key='secret',
+                                  region='region_123')
+        self.assertEqual(connection.auth_region_name, 'region_123')
+        self.assertEqual(connection.aws_access_key_id, 'access_key')
+        self.assertEqual(connection.aws_secret_access_key, 'secret')
+
+    def test_signing(self):
+        connection = ESConnection('mockservice.cc-zone-1.amazonaws.com',
+                                  aws_access_key_id='my_access_key',
+                                  aws_secret_access_key='secret',
+                                  region='region_123')
+        # create a request and sign it
+        request = connection.build_base_http_request('GET', '/', None)
+        request.authorize(connection)
+
+        # confirm the header contains signing method and key id
+        auth_header = request.headers['Authorization']
+        self.assertTrue('AWS4-HMAC-SHA256' in auth_header)
+        self.assertTrue('my_access_key' in auth_header)
+
+    def test_timeout(self):
+        def fake_connection(_address):
+            raise socket.timeout('fake error')
+        socket.create_connection = fake_connection
+        connection = ESConnection('mockservice.cc-zone-1.amazonaws.com',
+                                  aws_access_key_id='access_key',
+                                  aws_secret_access_key='secret',
+                                  region='region_123')
+        connection.num_retries = 0
+        with self.assertRaises(socket.error):
+            connection.make_request('GET', 'https://example.com')
+
+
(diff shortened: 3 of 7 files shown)


analytics_data_api/utils.py
```diff
@@ -47,49 +45,6 @@ def matching_tuple(answer):
     )
 
 
-def consolidate_answers(problem):
```
dsjen: I moved this into `views/problems.py`.  It both seemed odd in utils, and I was running into some very strange test failures in non-related db tests when importing `utils.py` into `apps.py` when this method remained here.

requirements/base.txt
```diff
@@ -1,10 +1,12 @@
+boto==2.22.1                        # MIT
```
dsjen: `boto` is the only new package included.  I alphabetized the packages.

analytics_data_api/v0/views/problems.py
```diff
@@ -55,6 +56,48 @@ class ProblemResponseAnswerDistributionView(generics.ListAPIView):
     serializer_class = ConsolidatedAnswerDistributionSerializer
     allow_empty = False
 
+    @classmethod
```
dsjen: This is moved here from `utils.py`.

analytics_data_api/utils.py
```diff
@@ -98,3 +53,10 @@ def dictfetchall(cursor):
         dict(zip([col[0] for col in desc], row))
         for row in cursor.fetchall()
     ]
+
+
+def class_for_name(path):
+    """ Returns the class given the full path. """
```
dan-f: Could you clarify what `path` is expected to be? For example, it seems like you're expecting a fully-qualified class name. I usually associate 'path' with a slash-separated directory or URL structure.

dan-f: Additionally, this function could give you any definition (class, variable, function, etc) instance given a fully-qualified definition. Maybe you could rename this to something like "load_fully_qualified_definition"?

analytics_data_api/v0/tests/test_connections.py
```diff
@@ -0,0 +1,35 @@
...
+    def test_region(self):
+        connection = ESConnection('mockservice.cc-zone-1.amazonaws.com',
+                                  aws_access_key_id='access_key',
+                                  aws_secret_access_key='secret')
+        self.assertEqual(connection.auth_region_name, 'region_123')
+
+
+@override_settings(ELASTICSEARCH_AWS_ACCESS_KEY_ID='access_key')
+@override_settings(ELASTICSEARCH_AWS_SECRET_ACCESS_KEY='secret')
+class BotoHttpConnectionTests(TestCase):
```
dan-f: Could you test that requests are actually signed by boto?

analytics_data_api/v0/connections.py
```diff
@@ -0,0 +1,69 @@
...
+
+class BotoHttpConnection(Connection):
+    """
+    Uses AWS configured connection to sign requests before they're sent to
+    elasticsearch nodes.
+    """
+
+    connection = None
+
+    def __init__(self, host='localhost', port=443, **kwargs):
```
dan-f: Why localhost 443?

analytics_data_api/v0/connections.py
```diff
@@ -0,0 +1,69 @@
...
+        super(BotoHttpConnection, self).__init__(host=host, port=port, **kwargs)
+        connection_params = {'host': host, 'port': port}
+
+        # Set the credentials through settings.  If not provided, boto will attempt to use
+        # default environment variables.
```
dan-f: What are the default variables for `aws_access_key_id` and `aws_secret_access_key`?

analytics_data_api/v0/connections.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        region = kwargs.pop('region', settings.ELASTICSEARCH_CONNECTION_DEFAULT_REGION)
+        kwargs.setdefault('is_secure', True)
+        super(ESConnection, self).__init__(*args, **kwargs)
+        self.auth_region_name = region
+        self.auth_service_name = 'es'
+
+    def _required_auth_capability(self):
```
dan-f: Could you document this method? I'm having trouble finding what it does.

analytics_data_api/v0/connections.py
```diff
@@ -0,0 +1,69 @@
...
+        if not isinstance(body, basestring):
+            body = json.dumps(body)
+        start = time.time()
+        response = self.connection.make_request(method, url, params=params, data=body)
+        duration = time.time() - start
+        raw_data = response.read()
+
+        # raise errors based on http status codes and let the client handle them
+        if not (200 <= response.status < 300) and response.status not in ignore:
+            self.log_request_fail(method, url, body, duration, response.status)
```
dan-f: Is this what the default `Connection` class does with HTTP error codes?

analytics_data_api/v0/tests/test_connections.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @patch('analytics_data_api.v0.connections.ESConnection.make_request')
+    def test_perform_request_success(self, mock_response):
+        mock_response.return_value.status = 200
+        connection = BotoHttpConnection()
+        status, _header, _data = connection.perform_request('get', 'http://example.com')
+        self.assertEqual(status, 200)
```
dan-f: We should also test that we logged success.

analytics_data_api/v0/tests/test_connections.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @patch('analytics_data_api.v0.connections.ESConnection.make_request')
+    def test_perform_request_error(self, mock_response):
+        mock_response.return_value.status = 500
+        connection = BotoHttpConnection()
+        with self.assertRaises(ElasticsearchException):
+            connection.perform_request('get', 'http://example.com')
```
dan-f: We should also test that we logged failure.

dan-f:
Looking good, but I'd like to see some more depth & breadth in the unit tests. @mulby mentioned that he didn't think this tool was production ready yet last time we chatted. I can't remember what the concerns were, but I'd like to make sure they're covered by test cases.

analytics_data_api/v0/connections.py
dsjen: No particular reason.  I can change that to 8080 to be less confusing.

dan-f: Ah I see. I guess it should be whatever the default AWS port is in this case.

analytics_data_api/v0/connections.py
dsjen: If any of these are `None`, boto will fall back to analogous environment variables.  When the environment variables aren't found, Boto will throw a `boto.exception.NoAuthHandlerFound` error.

analytics_data_api/v0/connections.py
dsjen: Yes, it's basically the same as [elasticsearch/connection/http_requests.py#L76](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/http_requests.py#L76).

analytics_data_api/v0/connections.py
dsjen: Sure thing!

analytics_data_api/v0/connections.py
dsjen: Very good point!  I'll set it to `9200`.

analytics_data_api/utils.py
dsjen: I see.  I'll update this to `load_fully_qualified_definition` and rename the variable to `definition`.

analytics_data_api/v0/connections.py
dan-f: Great!

analytics_data_api/v0/tests/test_connections.py
dsjen: I'm feeling a bit torn about writing tests for the logging.  `elasticsearch-py` [tests that functionality](https://github.com/elastic/elasticsearch-py/blob/master/test_elasticsearch/test_connection.py#L148) and testing it on our end end feels redundant and doesn't place trust in the library.  At the same time, the logging is a side effect of what it is that we're calling explicitly in our method, and thus arguably should be part of our tests.  I'm leaning to leaving it untested, but am interested in your thoughts.

dan-f: The `elasticsearch-py` tests don't run in our builds, let alone with our `BotoHttpConnection` connection class, though, so if we don't write our own test, that functionality will never be checked.

analytics_data_api/v0/connections.py
mulby: AWS elasticsearch service exposes the service on 443 so that you can use HTTPS URLs without specifying a port:

https://search.lkueroi73897.aws.com/

Instead of:

https://search.lkueroi73897.aws.com:9200/

Given that you would only use this connection class for aws elasticsearch clusters, I defaulted the port to 443.

analytics_data_api/v0/connections.py
mulby: We should consider accepting these variables as parameters to the constructor instead of reading them directly out of settings.

My pipeline code will not be able to import django.conf.settings.

analytics_data_api/v0/tests/test_connections.py
```diff
@@ -0,0 +1,35 @@
+from django.test import TestCase
+from django.test.utils import override_settings
+from elasticsearch.exceptions import ElasticsearchException
+from mock import patch
+
+from analytics_data_api.v0.connections import BotoHttpConnection, ESConnection
+
+
+class ESConnectionTests(TestCase):
```
mulby: I wonder if we shouldn't test the timeout case as well? Not sure what boto does when there is a timeout... but we should emulate it and make sure this connection code handles the case properly. Or perhaps all we need is a generic test that makes sure this "does the right thing" when an unexpected exception occurs in boto's request logic.

analytics_data_api/v0/views/problems.py
mulby: I love improvements like this! Thanks for taking a moment to leave the code better than you found it.

analytics_data_api/v0/connections.py
dan-f: Cool, 443 it is :)

analytics_data_api/v0/connections.py
dan-f: :+1: on Gabe's idea. Cleaner interface too, IMO

analytics_data_api/v0/connections.py
dsjen: Great, thanks for the info!

analytics_data_api/v0/connections.py
dsjen: We don't instantiate the connection.  Rather, the class is passed into elasticsearch-dsl's connection pool, thus changing the modifying the constructor won't get us what we want without modifications to elasticsearch-dsl.

The fallback if the settings aren't provided is that the environment variables are used.  @mulby -- does that work for the pipeline?  I suppose we'll need to conditionally import `django.conf.settings`.

Another way to do this might be to set the environment variable as defined in the django settings.  That feels weird though.  I'm open to other ideas!

mulby: I was hoping that you could pass arbitrary keyword arguments into create_connection and they would be passed along to the Connection class when it actually *is* instantiated.

Something like:

```python
connection_params = {'hosts': [settings.ELASTICSEARCH_LEARNERS_HOST]}
if settings.ELASTICSEARCH_CONNECTION_CLASS:
           connection_params['connection_class'] = class_for_name(settings.ELASTICSEARCH_CONNECTION_CLASS)
if settings.ELASTICSEARCH_AWS_ACCESS_KEY_ID:
           connection_params['aws_access_key_id'] = settings.ELASTICSEARCH_AWS_ACCESS_KEY_ID
if settings.ELASTICSEARCH_AWS_SECRET_ACCESS_KEY:
          connection_params['aws_secret_access_key'] = settings.ELASTICSEARCH_AWS_SECRET_ACCESS_KEY
           connections.connections.create_connection(**connection_params)
```

If that doesn't work, I'm OK with just wrapping the import in a try-except block that catches the ImportError.

dsjen: Unfortunately, that didn't work, but I've been digging into the elasticsearch client codebase and think I'm coming across some promising comments at the very least.

I can make this work!

dsjen:
@dan-f @mulby -- I've finally pushed an updated to address your comments.  Most of the changes were in the test code, where I added a check for the request signing, verifying logging, and timeout.  Check it out when you get the chance.  I now know more about boto than I ever wanted to.

requirements/base.txt
```diff
@@ -1,10 +1,12 @@
...
-Markdown==2.6    					# BSD
 django-model-utils==2.2 			# BSD
 djangorestframework==2.4.4			# BSD
-ipython==2.4.1						# BSD
 django-rest-swagger==0.2.8  		# BSD
 djangorestframework-csv==1.3.3		# BSD
 django-countries==3.2               # MIT
-elasticsearch-dsl==0.0.9           # Apache 2.0
+elasticsearch-dsl==0.0.9            # Apache 2.0
+ipython==2.4.1						# BSD
```
mulby: I assume ipython is here for debug purposes? Should it be in base.txt?

requirements/base.txt
```diff
@@ -1,10 +1,12 @@
...
+Markdown==2.6    					# BSD
```
mulby: same for markdown?

mulby:
:+1:

not a blocker, but did we decide to just copy this connection class between the repos?

requirements/base.txt
dsjen: That does seem out of place--not sure how that made it in there.  I'll remove it.

requirements/base.txt
dsjen: I'm going to get rid of that one too.  My guess is that at some point we wanted to support markdown in our swagger docs, but it's not configured.

analytics_data_api/v0/tests/test_connections.py
```diff
@@ -0,0 +1,66 @@
...
+                                  aws_secret_access_key='secret',
+                                  region='region_123')
+        # create a request and sign it
+        request = connection.build_base_http_request('GET', '/', None)
+        request.authorize(connection)
+
+        # confirm the header contains signing method and key id
+        auth_header = request.headers['Authorization']
+        self.assertTrue('AWS4-HMAC-SHA256' in auth_header)
+        self.assertTrue('my_access_key' in auth_header)
```
dan-f: Sorry to nitpick right when this is crossing over the line, but is it important to verify the exact structure of this request header? I'm assuming that `request.headers['Authorization']` is a string, and if so, would we just want to match that exact string?

dan-f:
One nitpick, but :+1: nevertheless. Thanks for incorporating my feedback!

analytics_data_api/v0/tests/test_connections.py
dsjen: That's a valid question, and yes, the `request.headers['Authorization']` is a string.  Verifying the exact encoding requires signing the request, and I could use boto for that, but then I'm just testing that boto is returning the same output that boto is doing--doesn't feel worthwhile.  I figure there are two key components to the signature, the encoding method ("AWS4-HMAC-SHA256") and the access key id ("my_access_key").

dsjen:
@mulby -- I'm thinking that we should just copy them for now.  The discussion on whether other teams will use it is ongoing.  If figure (with the help of the review), that we could pull this out into a package with minimal effort.

PR #105, opened 2016-01-14
Return null for empty learner fields.

dsjen:
@dan-f

### DIFF
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -332,6 +332,14 @@ class LearnerSerializer(serializers.Serializer, DefaultIfNoneMixin):
     enrollment_date = serializers.DateField(source='enrollment_date', format=settings.DATE_FORMAT)
     cohort = serializers.CharField(source='cohort')
 
+    def transform_segments(self, _obj, value):
+        # returns null instead of empty strings
+        return value or []
+
+    def transform_cohort(self, _obj, value):
+        # returns null instead of empty strings
+        return value or None
+
     def get_account_url(self, obj):
         if settings.LMS_USER_ACCOUNT_BASE_URL:
             return urljoin(settings.LMS_USER_ACCOUNT_BASE_URL, obj.username)
--- a/analytics_data_api/v0/tests/views/test_learners.py
+++ b/analytics_data_api/v0/tests/views/test_learners.py
@@ -118,25 +118,32 @@ class LearnerAPITestMixin(object):
             enrollment_date='2015-01-28',
     ):
         """Create a single learner roster entry in the elasticsearch index."""
+        body = {
+            'username': username,
+            'course_id': course_id,
+            'name': name if name is not None else username,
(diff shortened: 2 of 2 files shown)


analytics_data_api/v0/serializers.py
```diff
@@ -332,6 +332,14 @@ class LearnerSerializer(serializers.Serializer, DefaultIfNoneMixin):
     enrollment_date = serializers.DateField(source='enrollment_date', format=settings.DATE_FORMAT)
     cohort = serializers.CharField(source='cohort')
 
+    def transform_segments(self, _obj, value):
+        # returns null instead of empty strings
+        return value or None
+
+    def transform_cohort(self, _obj, value):
+        # returns null instead of empty strings
+        return value or None
```
dan-f: We need to support the concepts of "This user belongs to no cohort" and "This user belongs to the cohort named <the-empty-string>". It used to be that both were represented by the empty string value, but I'm now afraid that both are represented by the `null` value. I think that when a user belongs to no cohort (field is missing in ES), the API should return `null`, and when they belong to a cohort with the name being the-empty-string, we should return the empty string.

dsjen: I can definitely add in a check for an empty string cohort.  Why not?

Out of curiosity, is it a valid cohort name?  It's an odd use case to support, especially when filtering.

dsjen: It turns out that the field defaults to `''` when `null`.  We could create extend the elasticsearch dsl's `String` field type.  However, I'm not sure that I see the use case for an empty string cohort.  Does that happen?

dan-f: Looks like the empty string cohort does not actually happen. At least, the UI blocks you from entering a cohort with an empty name. Sorry for the confusion!

Given that, I like how this makes things work. `null` means the student belongs to no cohort, while any string means the student belongs to a cohort with that string as the name.

analytics_data_api/v0/serializers.py
```diff
@@ -332,6 +332,14 @@ class LearnerSerializer(serializers.Serializer, DefaultIfNoneMixin):
     enrollment_date = serializers.DateField(source='enrollment_date', format=settings.DATE_FORMAT)
     cohort = serializers.CharField(source='cohort')
 
+    def transform_segments(self, _obj, value):
+        # returns null instead of empty strings
+        return value or None
```
dan-f: I actually thought that we were getting the desired behavior with segments, which should be a list of strings. If the user belongs to no segments, I'd expect an empty list, not `null`. What do you think?

dan-f:
:+1: pending my comment about segments.

analytics_data_api/v0/serializers.py
dsjen: Thanks!  These are great questions.

analytics_data_api/v0/serializers.py
dsjen: Makes sense--will update shortly.

PR #107, opened 2016-03-22
Update primary keys for engagement range table.

dsjen:
@dan-f -- this is needed because the result store doesn't create an `id` column for the engagement ranges table, which is django's default primary key.


dsjen:
Drats...not sure why it's failing.  I'll investigate.

analytics_data_api/v0/models.py
```diff
@@ -445,7 +445,9 @@ class ModuleEngagementMetricRanges(models.Model):
     low_value and high_value.
     """
 
-    course_id = models.CharField(db_index=True, max_length=255)
+    # primary key is actually course_id, metric, and range_type, but only one
+    # primary is allowed in django
+    course_id = models.CharField(primary_key=True, db_index=True, max_length=255)
```
dan-f: I'm curious about the pros and cons to declaring `course_id` the primary key. We tend to use surrogate keys (default in django) everywhere else. I think the result store may be a special case due to its write patterns, but it's still worth considering. In particular, we should be confident that we'll never _ever_ need to change the `course_id` of a record in-between runs of the pipeline, because it has the potential to de-normalize the database. I don't see `primary_key=True` on any other model fields in this file, which leads me to believe that tables are always generated with an `id` field. @mulby do you have any thoughts here? Could we just generate this table with an `id` field?

mulby: It is actually setup this way precisely because of the special properties of this type of system.

We are trading flexibility (which we will likely never need) for query performance (which we know we want).

The pipeline currently provides the following assurances:

* Any table can be regenerated from scratch at any point in time.
* The API is a read-only consumer of the database.

Given those invariants, we can make performance optimizations like this one. Frankly, the fact that we didn't do this with all of our other tables is somewhat of an oversight. Using a meaningful primary key like this dramatically increases query performance at the cost of inflexibility and write performance. Given that we write infrequently and read often I think this is the correct trade-off.

Django assumes your database is a traditional OLTP system. This database is really just a cache, so we can (and should) take advantage of those differences in usage patterns.

mulby:
I'm surprised django doesn't provide a compound primary key mechanism... I'm a little wary of the differences in schema that the ORM thinks it's working with vs. the reality.

analytics_data_api/v0/models.py
dan-f: Thanks for that thorough explanation!

dsjen:
FYI, it looks like [MultipleColumnPrimaryKeys](https://code.djangoproject.com/wiki/MultipleColumnPrimaryKeys) has been "accepted/assigned and being worked on."  Maybe something will be ready for django 1.9!

PR #109, opened 2016-04-08
Include days with no activity in the engagement timeline

dan-f:
[AN-6960](https://openedx.atlassian.net/browse/AN-6960)

@dsjen please review

### DIFF
--- a/analytics_data_api/v0/models.py
+++ b/analytics_data_api/v0/models.py
@@ -1,3 +1,4 @@
+import datetime
 from itertools import groupby
 
 from django.conf import settings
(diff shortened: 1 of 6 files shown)


analytics_data_api/v0/models.py
```diff
@@ -418,7 +420,29 @@ def get_timelines(self, course_id, username):
                 day[engagement_type.name] = day.get(engagement_type.name, 0) + count_delta
             timelines.append(day)
 
-        return timelines
+        # Fill in dates that may be missing, since the result store doesn't
+        # store empty engagement entries.
+        full_timeline = []
+        default_timeline_entry = {
+            u'problems_attempted': 0,
```
dsjen: Nit: These could be generated by `EngagementType` instead.

dsjen:
Beautiful!  I have one small suggestion (take it or leave it).  :+1:

analytics_data_api/v0/models.py
dan-f: Yeah, I didn't like this being declared here. Good call!

PR #110, opened 2016-04-08
Learner analytics

dan-f:
[AN-6954](https://openedx.atlassian.net/browse/AN-6954)

@dsjen


dan-f:
Sad that coverage decreased. @dsjen are you familiar with coveralls? Perhaps we can find the lines of code that were missed and make the call whether they're worth testing or not.

dsjen:
That is too bad!  I'm not too familiar with coveralls unfortunately.  It does look like https://coveralls.io/builds/5719892 dissects it a bit more.  It looks like better coverage of utils might do the trick.

dsjen:
It's my sense that coverage is pretty good still, so this LGTM still.  It's up to you!

dan-f:
I'm going to check the coverage issue tomorrow morning, but I'll timebox it. I don't understand how coverage for `utils.py` is reported down 50% when it clearly increased on the branch...

dan-f:
^ 👏  ^

dsjen:
:+1: Wonderful!

PR #112, opened 2016-04-11
Fix the way we measure coverage

dan-f:
@dsjen FYI I messed up the rebase of learner-analytics, so I have to merge this in once more. I'll merge once tests pass.


dsjen:
Ok! :+1:

PR #113, opened 2016-04-11
Moved default load data options to make_option.

dsjen:
This sets the default options to the configuration of the management command.  It oddly fixes a problem where the values weren't being parsed.

@dan-f -- please review.


dan-f:
:+1:

PR #115, opened 2016-05-03
New API method to return tags distribution info

dmitry-viskov:
This PR is related to https://github.com/edx/edx-analytics-pipeline/pull/247 this one.

New API method:
```
curl "http://127.0.0.1:8100/api/v0/courses/foo-bar/problems_and_tags/" \
--header "Authorization: Token changeme" \
--header "Accept: application/json" -v
```
The structure of JSON answer:
```
[
    {
        "module_id": "foo-bar@problem+block@4c218e31d2014ba4a33f5cbc65c185b9",
        "total_submissions": 2,
        "correct_submissions": 1,
        "tags":
        {
            "difficulty": "Hard",
            "learning_outcome": "Learned nothing"
        }
    },
    {
        "module_id": "foo-bar@problem+block@68cf3df029324fa4a3736dee55357ec7",
        "total_submissions": 2,
        "correct_submissions": 0,
        "tags":
        {
            "difficulty": "Hard",
            "learning_outcome": "Learned a few things"
        }
    },
    ...
]
```

@mulby could you please take a look.

### DIFF
--- a/analytics_data_api/v0/views/courses.py
+++ b/analytics_data_api/v0/views/courses.py
@@ -628,9 +628,9 @@ class ProblemsListView(BaseCourseView):
         Returns a collection of submission counts and part IDs for each problem. Each collection contains:
 
             * module_id: The ID of the problem.
-            * total_submissions: Total number of submissions
+            * total_submissions: Total number of submissions.
             * correct_submissions: Total number of *correct* submissions.
-            * part_ids: List of problem part IDs
+            * part_ids: List of problem part IDs.
     """
     serializer_class = serializers.ProblemSerializer
     allow_empty = False
@@ -687,6 +687,53 @@ GROUP BY module_id;
                 row['created'] = datetime.datetime.strptime(created, '%Y-%m-%d %H:%M:%S')
 
         return rows
+
+
+# pylint: disable=abstract-method
+class ProblemsAndTagsListView(BaseCourseView):
+    """
+    Get the problems with the connected tags.
+
+    **Example request**
+
+        GET /api/v0/courses/{course_id}/problems_and_tags/
+
+    **Response Values**
+
+        Returns a collection of submission counts and tags for each problem. Each collection contains:
+
+            * module_id: The ID of the problem.
+            * total_submissions: Total number of submissions.
+            * correct_submissions: Total number of *correct* submissions.
+            * tags: Dictionary that contains pairs "tag key: tag value".
+    """
+    serializer_class = serializers.ProblemsAndTagsSerializer
+    allow_empty = False
+    model = models.ProblemsAndTags
+
+    def get_queryset(self):
+        queryset = self.model.objects.filter(course_id=self.course_id)
+        items = queryset.all()
+
+        result = {}
+
+        for v in items:
+            if v.module_id in result:
+                result[v.module_id]['tags'][v.tag_name] = v.tag_value
+                if result[v.module_id]['created'] < v.created:
+                    result[v.module_id]['created'] = v.created
+            else:
+                result[v.module_id] = {
+                    'module_id': v.module_id,
+                    'total_submissions': v.total_submissions,
+                    'correct_submissions': v.correct_submissions,
+                    'tags': {
+                        v.tag_name: v.tag_value
+                    },
+                    'created': v.created
+                }
+
+        return result.values()
 
 
 class VideosListView(BaseCourseView):
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,3 +6,4 @@ Ed Zarecor <ed@edx.org>
 Gabe Mulley <gabe@edx.org>
 Jason Bau <jbau@stanford.edu>
 John Jarvis <jarv@edx.org>
+Dmitry Viskov <dmitry.viskov@webenterprise.ru>
--- a/analytics_data_api/management/commands/generate_fake_course_data.py
+++ b/analytics_data_api/management/commands/generate_fake_course_data.py
@@ -225,6 +225,34 @@ class Command(BaseCommand):
                     course_id=course_id, start_date=start_date, end_date=end_date, metric=metric,
                     range_type='high', low_value=high_floor, high_value=max_value)
 
+    def generate_tags_distribution_data(self, course_id):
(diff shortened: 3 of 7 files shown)


edx-webhook:
Thanks for the pull request, @StraNNiKK! I've created [OSPR-1255](https://openedx.atlassian.net/browse/OSPR-1255) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

If you like, you can add yourself to the [AUTHORS](https://github.com/edx/edx-analytics-data-api/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

StraNNiKK:
jenkins run all

StraNNiKK:
Seems that something wrong with travis.
The previous travis tests were OK:
https://travis-ci.org/edx/edx-analytics-data-api/builds/127489842
before I've changed `AUTHORS` txt file

dsjen:
@StraNNiKK -- I'm rerunning the travis build to see if the error goes away.

dsjen:
@StraNNiKK -- tests have passed!

We've scheduled reviewing this PR for our next sprint, which starts in ~2 weeks.  I did a quick pass over it and it seems very straightforward.  Adding in a method for generating fake data would be helpful, similar to what is in in [generate_fake_course_data.py](https://github.com/edx/edx-analytics-data-api/blob/master/analytics_data_api/management/commands/generate_fake_course_data.py).

dsjen:
@StraNNiKK -- just wanted to let you know that reviewing this is pushed to our next sprint as it looks like https://github.com/edx/edx-analytics-pipeline/pull/247 is a prerequisite for this PR.

StraNNiKK:
@dsjen ok. thanks

StraNNiKK:
@dsjen hi. https://github.com/edx/edx-analytics-pipeline/pull/247 is merged so probably this PR could be reviewed

dsjen:
Thank you @dmitry-viskov!  I'm reviewing it and things are looking good.  One addition that I'd like to see is adding in some made up data to https://github.com/edx/edx-analytics-data-api/blob/master/analytics_data_api/management/commands/generate_fake_course_data.py.

analytics_data_api/v0/views/courses.py
```diff
@@ -689,6 +689,54 @@ def get_queryset(self):
...
+    model = models.ProblemsAndTags
+
+    def get_queryset(self):
+        queryset = self.model.objects.filter(course_id=self.course_id)
+        items = queryset.all()
+
+        result = {}
+
+        for v in items:
+            if v.module_id not in result:
```
dsjen: Nit: I find it a bit clearer if the `if` statements can be made in the positive.  My suggestion would be to change this to `if v.module_id in result:` likewise move the code blocks.  What do you think?

analytics_data_api/v0/views/courses.py
```diff
@@ -689,6 +689,54 @@ def get_queryset(self):
...
+    **Example request**
+
+        GET /api/v0/courses/{course_id}/problems_and_tags/
+
+    **Response Values**
+
+        Returns a collection of submission counts and tags for each problem. Each collection contains:
+
+            * module_id: The ID of the problem.
+            * tag_name: Name of the connected tag
```
dsjen: Just for consistency, please add a period to the end of the lines.

analytics_data_api/v0/views/courses.py
```diff
@@ -689,6 +689,54 @@ def get_queryset(self):
...
+            * tag_value: Value of the connected tag
```
dsjen: I think this description isn't quite right.  It looks like it's more like "tags" which is a map of tag names to values.

dsjen:
@mulby / @dmitry-viskov -- are the submission data in the tags_distribution table essentially the same that as returned in the other `/api/v0/courses/{course_id}/problems/`endpoint, which uses the `answer_distribution` table instead?  I think it could be confusing if the submission counts don't line up.

If they are the same, it seems like perhaps the tags could be moved into the original `.../problems/` endpoint.  Please let me know your thoughts!

dmitry-viskov:
@dsjen 

>are the submission data in the tags_distribution table essentially the same that as returned in the other /api/v0/courses/{course_id}/problems/endpoint, which uses the answer_distribution table instead?

From my point of view there are two reasons because of these I think it is better to use the separate API:
1. The returned values are different yet. The existed `problems` endpoint return all course problems but the new API should return only problems that are marked with any tag. If problem doesn't marked with any tag it should not be returned in this API. Also the other one thing is that the new functionality doesn't connected with `part_ids`.
2. Originally we with @mulby thought to add new field for the tag values into existed `answer_distribution` table. In this case the modification of the existed `problems` endpoint seems logical. But there were some circumstances because of which we decided not to do so. They are descried [here](https://github.com/edx/edx-analytics-pipeline/pull/247#issuecomment-215074770). So in case when we need to get data from the new table in my opinion it is better to use the separate API method (not to be confused).

analytics_data_api/v0/views/courses.py
dmitry-viskov: Yes. You are right. The positive condition looks better. I've changed this part of code

analytics_data_api/v0/views/courses.py
dmitry-viskov: done

analytics_data_api/v0/views/courses.py
dmitry-viskov: I've changed description a little

dmitry-viskov:
@dsjen hi. I've updated the PR and added the data generator into the `generate_fake_course_data.py` script.

dsjen:
@dmitry-viskov thank you so much for taking care of that.  I ran the command and it works perfectly.  I will merge this and thank you for this contribution!

dmitry-viskov:
@dsjen thank you for your help!
Could you please also review the related PR https://github.com/edx/edx-analytics-data-api-client/pull/24 ?
It was created to support the new API endpoint in the client library. It is pretty small.

PR #117, opened 2016-05-11
Added documentation to learner api for account url.

dsjen:
It looks like we forgot to add the `account_url` field to the api doc.

@dan-f -- please review.

### DIFF
--- a/analytics_data_api/v0/views/learners.py
+++ b/analytics_data_api/v0/views/learners.py
@@ -60,6 +60,7 @@ class LearnerView(LastUpdateMixin, CourseViewMixin, generics.RetrieveAPIView):
         Returns metadata and engagement data for the learner in JSON format.
 
             * username: The username of the enrolled learner.
+            * account_url:  URL to learner's account api endpoint.
             * enrollment_mode: The learner's selected learning track (for
               example, "audit" or "verified").
             * name: The learner's full name.


analytics_data_api/v0/views/learners.py
```diff
@@ -60,6 +60,7 @@ class LearnerView(LastUpdateMixin, CourseViewMixin, generics.RetrieveAPIView):
         Returns metadata and engagement data for the learner in JSON format.
 
             * username: The username of the enrolled learner.
+            * account_url:  URL to learner's account page.
```
dan-f: This is actually the URL to the learner's endpoint on the account API. There's probably a better way to word it than I just did, but I think it's important to distinguish between a page (e.g. HTML document) and API (e.g. JSON string).

dan-f:
Nice catch! I have a nitpick over wording, but 👍  otherwise!

analytics_data_api/v0/views/learners.py
dsjen: Oh, I actually thought that this should point to the learner's account page, e.g. https://courses.edx.org/u/[username].  I thought we were using this link in the learner's page on Insights.  It's good that we're clarifying this! 😄

dan-f: I can see how that would be valuable as well!

We will need the API endpoint as well in order to eventually fetch the profile image, bio, and other fields.

dan-f: how about something like `account_url` as well as `account_api_url`?

dsjen: I'm sort of wondering about whether these URLs should be in here at all now.  If Insights called out to the accounts API directly it could make the calls in parallel and decouple the sources of data a bit.  That way, if for some reason the data api went down, the application could (more) gracefully fail.

Thoughts?

dan-f: We had discussed something similar (except doing so on the analytics API) when designing this API. I think it's a reasonable idea, but the problem comes when fetching paginated sets of learners. To list N learners we'd need to send N network requests, which would be very slow.

dan-f: I could see us doing it for the single learner endpoint, but then there's an asymmetry between that and the list endpoint. Maybe that's alright?

dsjen:
Good conversation!  Documentation is updated and waiting on build to pass.

PR #118, opened 2016-05-13
Upgrade django debug toolbar

dan-f:
We don't use django debug toolbar, and it's causing issues in the Insights acceptance test suite.

@dsjen please review.


dsjen:
LGTM!  Something in the build is failing though.

dan-f:
@dsjen it's telling me the ES indexes already exist. I'm clearing the build cache and restarting.

dan-f:
Clearing the cache worked.

PR #119, opened 2016-05-19
Surface engagement metric ranges according to spec

dan-f:
[AN-6899](https://openedx.atlassian.net/browse/AN-6899)

Better handles edge cases when there's sparse data.

@dsjen please review.

Note to self: Update API Spec

### DIFF
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -412,28 +412,26 @@ class DateRangeSerializer(serializers.Serializer):
 
 class EnagementRangeMetricSerializer(serializers.Serializer):
     """
-    Serializes ModuleEngagementMetricRanges (low_range and high_range) into
-    the below_average, average, above_average ranges represented as arrays.
+    Serializes ModuleEngagementMetricRanges ('low', 'normal', and 'high') into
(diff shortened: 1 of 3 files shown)


analytics_data_api/v0/serializers.py
```diff
@@ -452,11 +450,17 @@ def get_engagement_ranges(self, obj):
         for entity_type in engagement_entity_types.AGGREGATE_TYPES:
             for event in engagement_events.EVENTS[entity_type]:
                 metric = '{0}_{1}'.format(entity_type, event)
+                # It's assumed that there may be any combination of low, normal,
```
dsjen: I thought we'd always have a "normal" range, no?

dsjen:
LGTM! 👍

analytics_data_api/v0/serializers.py
dan-f: Hmm. Let me chat with @mulby. I think in practice you're right, but that's defined at the pipeline/database layer rather than at the API.

dan-f: So yes, you're right. I'm going to change the test. I think what we decided is that whatever the DB exposes for ranges is what we expose in the API.

PR #123, opened 2016-06-23
Add DISTINCT to distinct_entity_count query

thallada:
This fixes half the issue of incorrect doubled counts on some engagement metrics. We should only count distinct entries (not duplicates) in the module_engagement table for the distinct_entity_count (the other half is the total_count sum, which is still double counts duplicates).

The pipeline is going to get a fix to remove duplicates that will make the `distinct=True` unnecessary, but since this is both easy and semantic it doesn't hurt to be defensive.


dsjen:
👍

PR #124, opened 2016-06-29
Run test ES server on different port to avoid conflict with edx-platform devstack

thallada:
I noticed that the edx-platform devstack was conflicting with the tests in this repo.

1. Start the devstack vagrant VM. The devstack uses version 0.90.13 of elasticsearch and binds to the default port of 9200 on the host machine.
2. Run `make test.run_elasticsearch` in this repo. Because port 9200 is taken, it will choose the next available port: 9201.
3. Run `make validate`. Because the settings/test.py specify that the elasticsearch server is on port 9200, it actually uses the elasticsearch server in the devstack. 0.90.13 doesn't support aggregation functions and most of the tests fail because of this.

My solution is to choose the rather arbitrary port 9223 to always run the test elasticsearch server on. If something is using that port, then the elasticsearch server will fail to start now.

@dsjen @ajpal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ test.install_elasticsearch:
 	echo "http.port: $(ELASTICSEARCH_PORT)" >> elasticsearch-$(ELASTICSEARCH_VERSION)/config/elasticsearch.yml
 
 test.run_elasticsearch:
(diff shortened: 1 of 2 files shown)


dsjen:
💯  👍

ajpal:
👍

PR #125, opened 2016-06-29
Enrollment tracks: add credit, do not merge audit & honor

thallada:
This updates our enrollment tracks on the API side to match what we are communicating elsewhere.

@dsjen, let me know if this is or isn't what you are expecting to work with in insights.

@dsjen @ajpal


analytics_data_api/management/commands/generate_fake_course_data.py
```diff
@@ -255,7 +256,7 @@ def handle(self, *args, **options):
         username = options['username']
         video_id = '0fac49ba'
         video_module_id = 'i4x-edX-DemoX-video-5c90cffecd9b48b188cbfea176bf7fe9'
-        start_date = datetime.datetime(year=2015, month=1, day=1, tzinfo=timezone.utc)
+        start_date = datetime.datetime(year=2016, month=1, day=1, tzinfo=timezone.utc)
```
dsjen: 🙌

dsjen:
Nice!  More code was deleted than added. :)  I tested this out and it looks great!  👍 

What do you think about holding off merging it until the learner analytics beta gets out in the next few days?

thallada:
@dsjen That sounds like a good idea. It would be nice to wait and synchronize this merge with an insights one that propagates these changes to the front-end too.

ajpal:
👍

PR #126, opened 2016-07-05
Add edx-django-release-util

efischer19:
WIP PR for [TE-1500](https://openedx.atlassian.net/browse/TE-1500)

I'm expecting tests to fail initially; the package has not been uploaded to PyPI just yet.


efischer19:
@dsjen Can you give this a quick review? It's adding tools that aren't used currently, but could help the platform/deployment team in the future.

analyticsdataserver/settings/base.py
```diff
@@ -202,6 +202,7 @@
 )
 
 THIRD_PARTY_APPS = (
+    'release_util',
```
dsjen: Nit: I'm not sure that it matters practically, but I think that the import order should be consistent between https://github.com/edx/edx-analytics-dashboard/pull/500 and this.  My slight preference is for it to go over the `rest_framework` imports.

dsjen:
Just one nit on import order. 👍

PR #127, opened 2016-07-25
Allows time component in start and end date parameters

thallada:
The [API docs](https://github.com/edx/edx-analytics-data-api/blob/master/analytics_data_api/v0/views/courses.py#L106) say that the date format for the course activity weekly endpoint is `YYYY-mm-ddTtttttt`, but the view was actually only accepting `YYYY-mm-dd`. All course API endpoints now support `YYYY-mm-ddTtttttt` (as well as `YYYY-mm-dd`).

@dsjen @ajpal

### DIFF
--- a/analytics_data_api/v0/tests/views/test_courses.py
+++ b/analytics_data_api/v0/tests/views/test_courses.py
@@ -130,20 +130,20 @@ class CourseViewTestCaseMixin(DemoCourseMixin):
 
     def assertIntervalFilteringWorks(self, expected_response, start_date, end_date):
         # If start date is after date of existing data, return a 404
-        date = (start_date + datetime.timedelta(days=30)).strftime(settings.DATE_FORMAT)
+        date = (start_date + datetime.timedelta(days=30)).strftime(settings.DATETIME_FORMAT)
         response = self.authenticated_get(
             '%scourses/%s%s?start_date=%s' % (self.api_root_path, self.course_id, self.path, date))
         self.assertEquals(response.status_code, 404)
 
         # If end date is before date of existing data, return a 404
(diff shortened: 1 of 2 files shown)


analytics_data_api/v0/views/courses.py
```diff
@@ -30,11 +30,17 @@ def get(self, request, *args, **kwargs):
         timezone = utc
 
         if start_date:
-            start_date = datetime.datetime.strptime(start_date, settings.DATE_FORMAT)
+            try:
```
dsjen: Nit: I guess both the `state_date` and `end_date` parsing could be refactored into a single method.

analytics_data_api/v0/tests/views/test_courses.py
```diff
@@ -130,20 +130,20 @@ def test_get_with_intervals(self):
...
+        date = (start_date - datetime.timedelta(days=30)).strftime(settings.DATETIME_FORMAT)
         response = self.authenticated_get(
             '%scourses/%s%s?end_date=%s' % (self.api_root_path, self.course_id, self.path, date))
         self.assertEquals(response.status_code, 404)
 
         # If data falls in date range, data should be returned
-        start_date = start_date.strftime(settings.DATE_FORMAT)
-        end_date = end_date.strftime(settings.DATE_FORMAT)
+        start_date = start_date.strftime(settings.DATETIME_FORMAT)
+        end_date = end_date.strftime(settings.DATETIME_FORMAT)
```
dsjen: Thanks for updating the tests!  I think for completion that a test for the `settings.DATE_FORMAT` case should be added.

dsjen:
Thank you for uncovering this!  I left one nit and a request for an additional test to cover the `settings.DATE_FORMAT` case.

👍

PR #131, opened 2016-08-03
Adds endpoint to download course reports

haikuginger:
This change creates a new endpoint at `/courses/{course_id}/reports/{report_name}/` that can be queried to create and provide a temporary downloadable link to an AWS-hosted CSV report.

- Currently, only the S3 storage architecture is supported as a host for these files
- Currently, only the `problem_response` report is supported as a download.

**Dependencies**: None

**Sandbox URL**: N/A

**Partner information**: 3rd party-hosted open edX instance

**Testing instructions**:

1. In an AWS S3 bucket, upload a report CSV. The filename (including path) should include the name of the report and the course ID of the course it was run on. E.g:

    `/course-v1:edX+DemoX+Demo_Course/problem_response.csv`

    Make a note of this location and bucket.

2. Make a note of your AWS access key ID and AWS secret access key

3. Determine the amount of time, in seconds, you'd like links to remain active after creation. In real-world use, this endpoint will be consumed by Javascript that than immediately downloads the file to the user's computer, so more than a few seconds to a minute shouldn't be necessary. In testing, you may want to allow yourself more time.

4. In an analytics devstack, checkout this branch and install requirements. This branch updates the version of `boto` required, and adds a requirement for `django-storages`.

5. Add the following to your `/edx/etc/analytics_api.yml` file:

    ```yaml
DEFAULT_FILE_STORAGE: storages.backends.s3boto.S3BotoStorage
AWS_ACCESS_KEY_ID: XXXXXXXXXXXXXX
AWS_SECRET_ACCESS_KEY: XXXXXXXXXXXXXX
AWS_STORAGE_BUCKET_NAME: report-download-bucket
COURSE_REPORT_FILE_LOCATION_TEMPLATE: '/{course_id}/{report_name}.csv'
COURSE_REPORT_DOWNLOAD_EXPIRY_TIME: 120
```
    Substitute your own AWS details for the above. For the `COURSE_REPORT_FILE_LOCATION_TEMPLATE` variable, note that this should be a Python format string; the endpoint will use this to locate the correct file for any given course/report combination. The `COURSE_REPORT_DOWNLOAD_EXPIRY_TIME` is given in seconds; update this to select whatever value you'd like - the default is two minutes.

6. Run the analytics API with `./manage.py runserver 0.0.0.0:8100 --insecure` or similar.

7. Using the Swagger console or `curl`, hit the endpoint and note that the response, among other details, provides you with a signed URL you can use to download the CSV file you uploaded directly. 

8. Note that when this link is used in a browser, it immediately downloads as a `text/csv` MIME type, and that the name of the downloaded file contains relevant information, such as the course ID, the name of the report, and the date the report was generated/last modified.

**Author notes and concerns**:

1. Right now, this endpoint supports a limited number of use cases, but there's no reason it can't be expanded to support other reports.

2. At the moment, the list of report(s) that can be downloaded is hardcoded, but it's a valid option to move this to a variable in `analytics_api.yml`.

**Reviewers**
- [x] @pomegranited
- [ ] @mulby

### DIFF
--- a/analytics_data_api/utils.py
+++ b/analytics_data_api/utils.py
@@ -1,8 +1,36 @@
 import datetime
 from importlib import import_module
+import re
 
 from django.db.models import Q
+from django.conf import settings
+from django.core.files.storage import default_storage
+from django.core.exceptions import SuspiciousFileOperation, SuspiciousOperation
 from rest_framework.authtoken.models import Token
+from opaque_keys.edx.locator import CourseKey
+from opaque_keys import InvalidKeyError
+
+from analytics_data_api.v0.exceptions import (
+    ReportFileNotFoundError,
+    CannotCreateReportDownloadLinkError
+)
+
+
+def get_filename_safe_course_id(course_id, replacement_char='_'):
+    """
+    Create a representation of a course_id that can be used safely in a filepath.
+    """
+    try:
+        course_key = CourseKey.from_string(course_id)
+        filename = unicode(replacement_char).join([course_key.org, course_key.course, course_key.run])
+    except InvalidKeyError:
+        # If the course_id doesn't parse, we will still return a value here.
+        filename = course_id
+
+    # The safest characters are A-Z, a-z, 0-9, <underscore>, <period> and <hyphen>.
+    # We represent the first four with \w.
+    # TODO: Once we support courses with unicode characters, we will need to revisit this.
+    return re.sub(r'[^\w\.\-]', unicode(replacement_char), filename)
 
 
 def delete_user_auth_token(username):
@@ -84,3 +112,121 @@ def date_range(start_date, end_date, delta=datetime.timedelta(days=1)):
     while cur_date < end_date:
         yield cur_date
         cur_date += delta
+
+
+def get_course_report_download_details(course_id, report_name):
+    """
+    Determine the path that the report file should be located at,
+    then return metadata sufficient for downloading it.
+    """
+    report_location_template = getattr(
+        settings,
+        'COURSE_REPORT_FILE_LOCATION_TEMPLATE',
+        '{course_id}_{report_name}.csv'
+    )
+    # Course IDs contain characters that may not be valid in various
+    # filesystems; here we remove them before looking for the file or
+    # creating the downloadable filename.
+    course_id = get_filename_safe_course_id(course_id)
+    report_location = report_location_template.format(
+        course_id=course_id,
+        report_name=report_name
+    )
+    try:
+        if not default_storage.exists(report_location):
+            raise ReportFileNotFoundError(course_id=course_id, report_name=report_name)
+    except (
+            AttributeError,
+            NotImplementedError,
+            ImportError,
+            SuspiciousFileOperation,
+            SuspiciousOperation
+    ):
+        # Error out if:
+        # - We don't have a method to determine file existence
+        # - Such a method isn't implemented
+        # - We can't import the specified storage class
+        # - We don't have privileges for the specified file location
+        raise CannotCreateReportDownloadLinkError
+
+    try:
+        last_modified = default_storage.modified_time(report_location)
+    except (NotImplementedError, AttributeError):
+        last_modified = None
+
+    try:
+        download_size = default_storage.size(report_location)
+    except (NotImplementedError, AttributeError):
+        download_size = None
+
+    download_filename = '{}-{}-{}.csv'.format(
+        course_id,
+        report_name,
+        # We need a date for the filename; if we don't know when it was last modified,
+        # use the current date and time to stamp the filename.
+        (last_modified or datetime.datetime.utcnow()).strftime('%Y%m%dT%H%M%SZ')
+    )
+    url, expiration_date = get_file_object_url(report_location, download_filename)
+
+    details = {
+        'course_id': course_id,
+        'report_name': report_name,
+        'download_url': url
+    }
+    # These are all optional items that aren't guaranteed. The URL isn't guaranteed
+    # either, but we'll raise an exception earlier if we don't have it.
+    if last_modified is not None:
+        details.update({'last_modified': last_modified.strftime(settings.DATETIME_FORMAT)})
+    if expiration_date is not None:
+        details.update({'expiration_date': expiration_date.strftime(settings.DATETIME_FORMAT)})
+    if download_size is not None:
+        details.update({'file_size': download_size})
+    return details
+
+
+def get_file_object_url(filename, download_filename):
+    """
+    Retrieve a download URL for the file, as well as a datetime object
+    indicating when the URL expires.
+
+    We need to pass extra details to the URL method, above and beyond just the
+    file location, to give us what we need.
+
+    This method supports S3 storage's optional response parameters that allow
+    us to set expiry time, as well as content disposition and content type
+    on any download made using the generated link.
+    """
+    # Default to expiring the link after two minutes
+    expire_length = getattr(settings, 'COURSE_REPORT_DOWNLOAD_EXPIRY_TIME', 120)
+    expires_at = get_expiration_date(expire_length)
+    try:
+        url = default_storage.url(
+            name=filename,
+            response_headers={
+                'response-content-disposition': 'attachment; filename={}'.format(download_filename),
+                'response-content-type': 'text/csv',
+                # The Expires header requires a very particular timestamp format
+                'response-expires': expires_at.strftime('%a, %d %b %Y %H:%M:%S GMT')
+            },
+            expire=expire_length
+        )
+    except TypeError:
+        # We got a TypeError when calling `.url()`; typically, this means that the arguments
+        # we passed aren't allowed. Retry with no extra arguments.
+        try:
+            url = default_storage.url(name=filename)
+            expires_at = None
+        except (AttributeError, TypeError, NotImplementedError):
+            # Another error, for unknown reasons. Can't recover from this; fail fast
+            raise CannotCreateReportDownloadLinkError
+    except (AttributeError, NotImplementedError):
+        # Either we can't find a .url() method, or we can't use it. Raise an exception.
+        raise CannotCreateReportDownloadLinkError
+    return url, expires_at
+
+
+def get_expiration_date(seconds):
+    """
+    Determine when a given link will expire, based on a given lifetime
+    """
+    return datetime.datetime.utcnow() + datetime.timedelta(seconds=seconds)
--- a/analytics_data_api/v0/urls/courses.py
+++ b/analytics_data_api/v0/urls/courses.py
@@ -14,7 +14,8 @@ COURSE_URLS = [
     ('enrollment/location', views.CourseEnrollmentByLocationView, 'enrollment_by_location'),
     ('problems', views.ProblemsListView, 'problems'),
     ('problems_and_tags', views.ProblemsAndTagsListView, 'problems_and_tags'),
-    ('videos', views.VideosListView, 'videos')
+    ('videos', views.VideosListView, 'videos'),
+    ('reports/(?P<report_name>[a-zA-Z0-9_]+)', views.ReportDownloadView, 'reports'),
 ]
 
 urlpatterns = []
--- a/analytics_data_api/v0/views/courses.py
+++ b/analytics_data_api/v0/views/courses.py
@@ -799,9 +799,7 @@ class ReportDownloadView(APIView):
             * expiration_date: The date through which the link will be valid
             * file_size: The size in bytes of the CSV download
     """
-    enabled_reports = (
-        'problem_response'
-    )
+    enabled_reports = settings.ENABLED_REPORT_IDENTIFIERS
 
     def get(self, _request, course_id, report_name):
         if report_name in self.enabled_reports:
--- a/analytics_data_api/v0/views/courses.py
+++ b/analytics_data_api/v0/views/courses.py
@@ -9,11 +9,14 @@ from django.db.models import Max
 from django.http import Http404
 from django.utils.timezone import make_aware, utc
 from rest_framework import generics
+from rest_framework.response import Response
+from rest_framework.views import APIView
 from opaque_keys.edx.keys import CourseKey
 
 from analytics_data_api.constants import enrollment_modes
-from analytics_data_api.utils import dictfetchall
+from analytics_data_api.utils import dictfetchall, get_course_report_download_details
 from analytics_data_api.v0 import models, serializers
+from analytics_data_api.v0.exceptions import ReportFileNotFoundError
 
 from analytics_data_api.v0.views.utils import raise_404_if_none
 
@@ -772,3 +775,37 @@ class VideosListView(BaseCourseView):
     def apply_date_filtering(self, queryset):
         # no date filtering for videos -- just return the queryset
         return queryset
+
+
+class ReportDownloadView(APIView):
+    """
+    Get information needed to download a CSV report
+
+    **Example request**
+
+        GET /api/v0/courses/{course_id}/reports/{report_name}/
+
+    **Response Values**
+
+        Returns a single object with data about the report, with the following data:
+
+            * course_id: The ID of the course
+            * report_name: The name of the report
+            * download_url: The Internet location from which the report can be downloaded
+
+        The object may also return these items, if supported by the storage backend:
+
+            * last_modified: The date the report was last updated
+            * expiration_date: The date through which the link will be valid
+            * file_size: The size in bytes of the CSV download
+    """
+    enabled_reports = (
+        'problem_response'
+    )
+
+    def get(self, _request, course_id, report_name):
+        if report_name in self.enabled_reports:
+            response = get_course_report_download_details(course_id, report_name)
+            return Response(response)
+        else:
+            raise ReportFileNotFoundError(course_id=course_id, report_name=report_name)
--- a/analyticsdataserver/settings/base.py
+++ b/analyticsdataserver/settings/base.py
@@ -177,6 +177,8 @@ MIDDLEWARE_CLASSES = (
     'analytics_data_api.v0.middleware.CourseNotSpecifiedErrorMiddleware',
     'analytics_data_api.v0.middleware.CourseKeyMalformedErrorMiddleware',
     'analytics_data_api.v0.middleware.ParameterValueErrorMiddleware',
+    'analytics_data_api.v0.middleware.ReportFileNotFoundErrorMiddleware',
+    'analytics_data_api.v0.middleware.CannotCreateDownloadLinkErrorMiddleware',
 )
 ########## END MIDDLEWARE CONFIGURATION
 
@@ -204,6 +206,7 @@ THIRD_PARTY_APPS = (
     'rest_framework.authtoken',
     'rest_framework_swagger',
     'django_countries',
+    'storages'
 )
 
 LOCAL_APPS = (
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

**Comment by edx-webhook** [comment]
Thanks for the pull request, @haikuginger! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-analytics-data-api&number=131

**Comment by edx-webhook** [comment]
Thanks for the pull request, @haikuginger! I've created [OSPR-1375](https://openedx.atlassian.net/browse/OSPR-1375) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

If you like, you can add yourself to the [AUTHORS](https://github.com/edx/edx-analytics-data-api/blob/master/AUTHORS)      file for this repo, though that isn't required.     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

File: analytics_data_api/utils.py [hunk 1]
```diff
@@ -84,3 +94,92 @@ def date_range(start_date, end_date, delta=datetime.timedelta(days=1)):
...
         yield cur_date
         cur_date += delta
+
+
+def get_course_report_download_details(course_id, report_name):
+    """
+    Determine the path that the report file should be located at,
+    then return metadata sufficient for downloading it.
+    """
+    report_location_template = getattr(settings, 'COURSE_REPORT_FILE_LOCATION_TEMPLATE')
```
pomegranited: To reduce the dependencies between this PR and your configuration PR (and to deal with the case where users simply update the analytics_api code without re-installing via configuration), we need to make sure that this code works even with the original analytics_api.yml file installed.

So this check needs to provide a default value, e.g.,

`report_location_template = getattr(settings, 'COURSE_REPORT_FILE_LOCATION_TEMPLATE', 'invalid-report-template')`

analytics_data_api/utils.py
```diff
@@ -84,3 +94,92 @@ def date_range(start_date, end_date, delta=datetime.timedelta(days=1)):
...
+    report_location = report_location_template.format(
+        course_id=course_id,
+        report_name=report_name
+    )
+    if getattr(settings, 'DEFAULT_FILE_STORAGE', None) != AWS_S3_FILE_STORAGE_CLASS:
```
pomegranited: Since you are only using methods specified by the Django Storage API, and nothing in the code is explicitly AWS-specific, I'd rather you removed this check.

analytics_data_api/utils.py
```diff
@@ -84,3 +94,92 @@ def date_range(start_date, end_date, delta=datetime.timedelta(days=1)):
...
+        # At present, we do not support linking from storage other than S3.
+        raise CannotCreateReportDownloadLinkError
+
+    if not default_storage.exists(report_location):
```
pomegranited: However, I do agree that it's important to check that django.storage is configured and available.  You can do this by replacing this stanza with:
```
    try:
        if not default_storage.exists(report_location):
            raise ReportFileNotFoundError(course_id=course_id, report_name=report_name)
    except (NotImplementedError, ImportError, SuspiciousFileOperation):
        raise CannotCreateReportDownloadLinkError
```

* `NotImplementedError` catches the case where there's a django.storage chosen that doesn't implement `exists()`
* `ImportError` catches the case where there's an invalid django.storage class configured
* `SuspiciousFileOperation` catches the case where no settings.DEFAULT_FILE_STORAGE is provided ([see above](#r73460844)), and so django.storage is using the default FileSystemStorage, and our report file sit outside the settings.MEDIA_DIR.

analytics_data_api/utils.py
```diff
@@ -84,3 +94,92 @@ def date_range(start_date, end_date, delta=datetime.timedelta(days=1)):
...
+
+    # Note that course IDs contain characters that aren't valid for filenames in
+    # Windows. However, we'll trust the client to save them with the appropriate
+    # alternate naming scheme (replace ':' with '_' and so on)
+    download_filename = '{}-{}-{}.csv'.format(
+        course_id,
+        report_name,
+        # We need a date for the filename; if we don't know when it was last modified,
+        # use the current date and time to stamp the filename.
+        (last_modified or datetime.datetime.utcnow()).strftime('%Y%m%dT%H%M%S')
```
pomegranited: Nit: could you append a 'Z' to the datetime format?  It's not ISO-8601 or anything, but at least gives a clue that the datetime is in UTC.

analytics_data_api/utils.py
```diff
@@ -84,3 +94,92 @@ def date_range(start_date, end_date, delta=datetime.timedelta(days=1)):
...
+
+
+def get_file_object_url(filename, download_filename):
+    """
+    Retrieve a download URL for the file, as well as a datetime object
+    indicating when the URL expires.
+
+    We need to pass extra details to the URL method, above and beyond just the
+    file location, to give us what we need.
+    """
```
pomegranited: Please add a comment indicating that this logic may need to be changed to support non-S3 django.storage modules.

analytics_data_api/utils.py
```diff
@@ -84,3 +94,92 @@ def date_range(start_date, end_date, delta=datetime.timedelta(days=1)):
...
+    # Default to expiring the link after one day
+    expire_length = getattr(settings, 'COURSE_REPORT_DOWNLOAD_EXPIRY_TIME', 86400)
```
pomegranited: As per my comment on the configuration PR, could you use a lower number for the default?  Something closer to the expected use case, e.g. 120.

analytics_data_api/v0/tests/test_connections.py
```diff
@@ -33,7 +33,7 @@ def test_signing(self):
         self.assertTrue('my_access_key' in auth_header)
 
     def test_timeout(self):
-        def fake_connection(_address):
+        def fake_connection(_address, _timeout):
```
pomegranited: Is this change necessary?

analytics_data_api/v0/urls/courses.py
```diff
@@ -14,7 +14,8 @@
     ('enrollment/location', views.CourseEnrollmentByLocationView, 'enrollment_by_location'),
     ('problems', views.ProblemsListView, 'problems'),
     ('problems_and_tags', views.ProblemsAndTagsListView, 'problems_and_tags'),
-    ('videos', views.VideosListView, 'videos')
+    ('videos', views.VideosListView, 'videos'),
+    ('reports/(?P<report_name>[a-zA-Z0-9_]+)', views.ReportDownloadView, 'reports')
```
pomegranited: nit: Please append a trailing comma, so future additions don't have to modify this line (like you had to do).

pomegranited:
@haikuginger This change looks awesome!  And it works as described.

I've left some comments, mainly around ensuring that the updated code works with an old analytics_api.yml file, and removing a bit of the forced S3-specificity, while leaving the comments and behaviour in place.

Please address those, and I'll give it a final check before approving.

Excellent work!  :)

analytics_data_api/v0/tests/test_connections.py
haikuginger: Yes; the upgraded Boto library passes two arguments here, so we get a TypeError when testing otherwise.

analytics_data_api/utils.py
haikuginger: Unfortunately, our use of `.url()` actually is AWS-specific - most other backends for `storages` don't support the arguments we're passing. We'd just get a TypeError lower down. Should I catch that instead?

analytics_data_api/utils.py
haikuginger: :+1:

analytics_data_api/utils.py
haikuginger: Add AttributeError, and you've got a party!

analytics_data_api/utils.py
haikuginger: :+1:

analytics_data_api/utils.py
haikuginger: :+1:

analytics_data_api/utils.py
haikuginger: 👍

analytics_data_api/v0/urls/courses.py
haikuginger: :+1:

analytics_data_api/utils.py
pomegranited: Ah, I got thrown by [files.storage.Storage.url()](https://docs.djangoproject.com/en/1.9/ref/files/storage/#django.core.files.storage.Storage.url).

Yes, could you please try/catch that one, and throw your custom "CannotCreateReportDownloadLinkError" exception on TypeError or NotImplementedError?

analytics_data_api/utils.py
pomegranited: Wheee!

analytics_data_api/v0/tests/test_connections.py
pomegranited: Gotcha.

haikuginger:
@pomegranited: Comments addressed!

pomegranited:
:+1:

- [x] I tested this using:
* an existing analytics_api.yml file without the above changes (defaults to django FileStorage, and 404s on the missing file)
* an analytics_api.yml file containing invalid AWS auth details
* an analytics_api.yml file updated as above
* invalid course_id
* invalid report_name
- [x] I read through the code
- [x] Includes documentation

haikuginger:
@mulby, this is ready for review.

pomegranited:
@haikuginger :+1 for the updated code.  Thank you for getting that in!

bradenmacdonald:
@haikuginger Is there any way to know which courses have the `problem_response` report available, other than calling this API for each course and checking if it returns 200 or 404? And is there any way to check without generating a temporary URL? (e.g. perhaps a `HEAD /api/v0/courses/{course_id}/reports/{report_name}/` could tell me if the report is available without actually downloading it)

haikuginger:
@bradenmacdonald: Addressing each question in turn.

> Is there any way to know which courses have the problem_response report available, other than calling this API for each course and checking if it returns 200 or 404?

I think the trick here is that the Analytics API is very loosely coupled to the actual EdX platform. You'll note that for all the other course APIs, the course ID must be specified. This is because we don't actually have a comprehensive list of courses to work with.

It's certainly possible to create an endpoint that could "discover" the existence of report files through pattern matching, but I'm not sure what the scope exactly would end up being - right now, the endpoint is pretty ignorant about filesystem structures; it just knows that if it formats a string with a couple variables, it can use that to generate a link if there's a file at that location. Expanding it is probably worth a story of its own.

> [Is] there any way to check without generating a temporary URL?

Sure. I'd just need to define a `.head()` method for `ReportDownloadView` that checks `default_storage.exists()` for the relevant location. That'd be pretty quick and easy to implement. Depending on the additional information you might want with it, though, it doesn't necessarily get us much compared to the plain `GET`; if you want useful metadata (last modified, size, etc), we'd have to make remote API calls to get those details anyway, and we'd just be saving the very last call.

Note that right now, the API itself isn't initiating a download of the CSV; it's just setting up a link and providing that link to the client. And, in production, we can make those links relatively short-lived, so there isn't necessarily a large security concern in regards to generating them.

bradenmacdonald:
@haikuginger Thanks for the comments; that makes sense to me.

analytics_data_api/utils.py
```diff
@@ -84,3 +115,113 @@ def date_range(start_date, end_date, delta=datetime.timedelta(days=1)):
...
+        'course_id': course_id,
+        'report_name': report_name,
+        'download_url': url
+    }
+    # These are all optional items that aren't guaranteed. The URL isn't guaranteed
+    # either, but we'll raise an exception earlier if we don't have it. Currently, support
+    # is only present for S3, which has all these data types, but this future-proofs
+    # us for scenarios where we might have a storage provider that doesn't support them.
+    if last_modified is not None:
+        details.update({'last_modified': last_modified.strftime(ISO_8601_FORMAT_STRING)})
```
bradenmacdonald: @haikuginger  API JSON responses like this should use `.strftime(settings.DATETIME_FORMAT)` for consistency.  Currently, this field is giving me an error when I try to parse it in Insights using the standard API client.

analytics_data_api/utils.py
```diff
@@ -1,8 +1,39 @@
...
 from rest_framework.authtoken.models import Token
+from opaque_keys.edx.locator import CourseKey
+from opaque_keys import InvalidKeyError
+
+from analytics_data_api.v0.exceptions import (
+    ReportFileNotFoundError,
+    CannotCreateReportDownloadLinkError
+)
+
+AWS_S3_FILE_STORAGE_CLASS = 'storages.backends.s3boto.S3BotoStorage'
```
pomegranited: @haikuginger Sorry I didn't pick this up, but could you remove this line too?  It's not required since we're no longer requiring the AWS storage class.

haikuginger:
All pending comments have been addressed.

pomegranited:
👍  Thanks for that, @haikuginger .

analytics_data_api/utils.py
```diff
@@ -84,3 +112,113 @@ def date_range(start_date, end_date, delta=datetime.timedelta(days=1)):
...
+        url = default_storage.url(
+            name=filename,
+            response_headers={
+                'response-content-disposition': 'attachment; filename={}'.format(download_filename),
+                'response-content-type': 'text/csv',
+                # The Expires header requires a very particular timestamp format
+                'response-expires': expires_at.strftime('%a, %d %b %Y %H:%M:%S GMT')
+            },
+            expire=expire_length
+        )
```
pomegranited: @haikuginger So, I know we said at the beginning of this task that we were only going to support S3 files, and you did that, which was great.

But @bradenmacdonald raised the point during his testing that he's never had to set up S3 storage for the analytics devstack before, and now @maxrothman is referring to [supporting the default filesystem storage](https://github.com/edx/configuration/pull/3286#issuecomment-239831593) in the configuration.

So I think we should support the default filesystem storage in dev, and allow S3 to be used in production.  

To help things along, I've tested the following code and configuration on my devstack, and it works, though of course doesn't have the nice expiry and timestamped download filename that S3 allows.

`analytics_data_api/url.py` diff:

```diff
-    Currently, this method only supports S3 storage, and will need to be changed to
-    support building URLs on other storage providers.
+    Currently, this method only supports FileSystemStorage and S3 storage, and 
+    will need to be changed to support building URLs on other storage providers.
     """
-    # Default to expiring the link after two minutes
-    expire_length = getattr(settings, 'COURSE_REPORT_DOWNLOAD_EXPIRY_TIME', 120)
-    expires_at = get_expiration_date(expire_length)
-    try:
-        url = default_storage.url(
-            name=filename,
+    url_params = dict(name=filename)
+    expires_at = None
+
+    expire_length = getattr(settings, 'COURSE_REPORT_DOWNLOAD_EXPIRY_TIME', None)
+    if expire_length is not None:
+        expires_at = get_expiration_date(expire_length)
+        url_params += dict(
             response_headers={
                 'response-content-disposition': 'attachment; filename={}'.format(download_filename),
                 'response-content-type': 'text/csv',
@@ -210,6 +216,8 @@ def get_file_object_url(filename, download_filename):
             },
             expire=expire_length
         )
+    try:
+        url = default_storage.url(**url_params)
     except (AttributeError, TypeError, NotImplementedError):
         # Either we can't find a .url() method, or we can't use it. Raise an exception.
         raise CannotCreateReportDownloadLinkError
```


configuration:  append to `analyticsdataserver/settings/base.py`

```yaml
REPORT_DOWNLOAD_BACKEND:
  DEFAULT_FILE_STORAGE: 'django.core.files.storage.FileSystemStorage'
  MEDIA_ROOT: '/edx/app/analytics_api/analytics_api/static/reports'
  MEDIA_URL: 'http://localhost:8100/static/reports/'   # note absolute URL, and trailing slash
  COURSE_REPORT_FILE_LOCATION_TEMPLATE: '{course_id}_{report_name}.csv'   # note no leading slash
```

pomegranited:
Excellent, thank you @haikuginger !  

:+1:

- [x] I tested this by running the updated code with both sets of configuration (FileStorage and S3), and ensuring the report files downloaded, and 404'd where appropriate.
- [x] I read through the code.
- [x] Includes documentation, and should be enough information for anyone needing to add additional storage mechanisms.

pomegranited:
@haikuginger  Could you please squash these commits, so I can cherry-pick this change for our client deployment?  Thank you!

haikuginger:
@pomegranited , rebased from [`4a17c7`](https://github.com/edx/edx-analytics-data-api/commits/4a17c731b19c4f6f7990bc248c66e4424da39bd8).

analyticsdataserver/settings/base.py
```diff
@@ -309,6 +312,12 @@
 # base url to generate link to user api
 LMS_USER_ACCOUNT_BASE_URL = None
 
+# storage settings for report downloads
+DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
+MEDIA_ROOT = '/edx/app/analytics_api/analytics_api/static/reports'
```
thallada: Can we use DJANGO_ROOT here instead? If we are running the server outside of the analytics devstack this will likely fail due to permissions or because the directory doesn't exist.

Something like: `normpath(join(DJANGO_ROOT, 'static', 'reports'))`

requirements/base.txt
```diff
@@ -1,5 +1,5 @@
-boto==2.22.1                        # MIT
 Django==1.8.14 						# BSD License
+boto==2.42.0                        # MIT
```
thallada: We recently upgraded to Django 1.9 along with some of these third-party packages, so this will conflict, but master should be running on the same version of boto now.

analytics_data_api/v0/tests/test_connections.py
thallada: I also encountered this when upgrading boto on master. This will conflict because I replaced the parameters to this function with `*args`. This also works, so feel free to keep these changes when resolving the conflict.

analytics_data_api/v0/views/courses.py
```diff
@@ -763,3 +766,37 @@ class VideosListView(BaseCourseView):
...
+
+        The object may also return these items, if supported by the storage backend:
+
+            * last_modified: The date the report was last updated
+            * expiration_date: The date through which the link will be valid
+            * file_size: The size in bytes of the CSV download
+    """
+    enabled_reports = (
+        'problem_response'
+    )
```
thallada: I know you mentioned this in your description, but I think we should at the very least move this tuple to a django setting. Then we could make that setting pull from `analytics_api.yml` in base.py.

analyticsdataserver/settings/base.py
```diff
@@ -309,6 +312,12 @@
...
+MEDIA_URL = 'http://localhost:8100/static/reports/'
```
thallada: Since we are overriding the original MEDIA_URL definition in this file, it's not needed and can be deleted or replaced (on line 96).

Also, the port number would vary in different dev environments, but I can't think of a good way to get that in the settings or at runtime. So, for now devs can just override this in local.py if they run their dev server on a different port.

haikuginger:
Rebased from c41091d.

analyticsdataserver/settings/base.py
haikuginger: I have no objection to that.

requirements/base.txt
haikuginger: ✅

analytics_data_api/v0/views/courses.py
haikuginger: Makes sense to me.

thallada:
Great! :+1: Thanks for the fast response :)

PR #135, opened 2016-08-29
Update DRF to 3.4.6

dsjen:
Bump the major version of django rest framework.  The types of changes include:
* `request.QUERY_PARAMS` removed and changed to `request.query_params`
* `DocType` date fields need format
* `serializers.SerializerMethodField()` no longer needs pointer to transform standard name
* `pagination.PaginationSerializer` updated to `pagination.PageNumberPagination`
* Updated lots of test expectations to unicode

I squashed a bunch of commits and tried to keep our individual contributions.  @thallada -- please review!

### DIFF
--- a/analytics_data_api/v0/models.py
+++ b/analytics_data_api/v0/models.py
@@ -234,7 +234,7 @@ class Video(BaseVideo):
 
 class RosterUpdate(DocType):
 
-    date = Date()
+    date = Date(format=settings.DATE_FORMAT)
 
     # pylint: disable=old-style-class
     class Meta:
@@ -265,8 +265,8 @@ class RosterEntry(DocType):
     attempt_ratio_order = Integer()
     discussion_contributions = Integer()
     videos_watched = Integer()
-    enrollment_date = Date()
-    last_updated = Date()
+    enrollment_date = Date(format=settings.DATE_FORMAT)
+    last_updated = Date(format=settings.DATE_FORMAT)
 
     # pylint: disable=old-style-class
     class Meta:
@@ -460,7 +460,7 @@ class ModuleEngagement(models.Model):
 
     course_id = models.CharField(db_index=True, max_length=255)
     username = models.CharField(max_length=255)
-    date = models.DateTimeField()
+    date = models.DateField()
     # This will be one of "problem", "video" or "discussion"
     entity_type = models.CharField(max_length=255)
     # For problems this will be the usage key, for videos it will be the html encoded module ID,
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -47,6 +47,7 @@ class ModelSerializerWithCreatedField(serializers.ModelSerializer):
     created = serializers.DateTimeField(format=settings.DATETIME_FORMAT)
 
 
+# pylint: disable=abstract-method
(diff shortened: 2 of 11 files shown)


analytics_data_api/v0/serializers.py
```diff
@@ -329,35 +332,47 @@ class Meta(object):
         )
 
 
+# pylint: disable=abstract-method
 class LastUpdatedSerializer(serializers.Serializer):
-    last_updated = serializers.DateField(source='date', format=settings.DATE_FORMAT)
+    last_updated = serializers.DateTimeField(source='date', format=settings.DATE_FORMAT)
```
thallada: Why is this a DateTimeField if we are only storing the date without time?

analytics_data_api/v0/tests/views/test_courses.py
```diff
@@ -210,10 +210,12 @@ def assertValidActivityResponse(self, activity_type, count):
 
     @staticmethod
     def get_activity_record(**kwargs):
+        # TODO: DJ -- I'm not sure why I have to format it...
```
thallada: I guess drf started to store datetimes as strings in the response.data instead of datetime objects.

analytics_data_api/v0/views/utils.py
```diff
@@ -10,3 +11,16 @@ def split_query_argument(argument):
         return argument.split(',')
     else:
         return None
+
+
+def raise_404_if_none(func):
+    """
+    Decorator for raiseing Http404 if function evaulation is falsey (e.g. empty queryset).
+    """
```
thallada: nice!

analytics_data_api/v0/serializers.py
thallada: Ok, I think I get it. So we store it as a datetime, but only ever display/serialize the date part?

thallada:
Just had one question. Looks great! :+1:

analytics_data_api/v0/tests/views/test_courses.py
dsjen: Right on!  I'll remove that TODO as well.

analytics_data_api/v0/serializers.py
dsjen: Oh yeah!  I remember the problem I ran into with this initially.  I tried using a `DateField`, but elasticsearch stores everything as a datetime, even if it really is just a date.  When using a `DateField`, I get a "coercion" error. 😞

analytics_data_api/v0/views/utils.py
dsjen: I had been dying to write a decorator! 😀

analytics_data_api/v0/models.py
```diff
@@ -460,7 +460,7 @@ class ModuleEngagement(models.Model):
 
     course_id = models.CharField(db_index=True, max_length=255)
     username = models.CharField(max_length=255)
-    date = models.DateTimeField()
+    date = models.DateField()
```
thallada: @dsjen could this have something to do with https://openedx.atlassian.net/browse/AN-7828 ? Only datetimes have the utcoffset function. Why did we have to change this field?

dsjen: Yep!  That's probably it!

PR #136, opened 2016-08-30
Fixed serialization of problem part IDs and tags.

dsjen:
I misunderstood what was happening during the DRF upgrade with serialization of lists.  This fixes the JSON so that the entire object isn't returned as a string.

Please review!


thallada:
Glad you caught this! Looks good :+1:

dsjen:
Always learning! 😄

PR #137, opened 2016-08-31
Upgrade to Django 1.9

thallada:
[AN-7227](https://openedx.atlassian.net/browse/AN-7227)

I also upgraded every python package I could to the latest versions that are compatible.


analytics_data_api/v0/tests/test_urls.py
```diff
@@ -6,7 +6,7 @@ class UrlRedirectTests(TestCase):
     api_root_path = '/api/v0/'
 
     def assertRedirectsToRootPath(self, path, **kwargs):
-        assert_kwargs = {'status_code': 301}
+        assert_kwargs = {'status_code': 302}
```
dsjen: Nice!

dsjen:
Thank you!  I tested this out via Insights too and everything runs.  👍

PR #138, opened 2016-09-01
Fix: add "--run-syncdb" to the migrate Make rule

thallada:
I missed this in the Django 1.9 upgrade: in 1.9, syncdb was turned off by default on the `migrate` command. This was preventing all of the analytics tables from being created in the analytics.db when it is first created (which happens in Travis every build).

We'll know this fix worked once we merge this, run the acceptance tests in insights again, and they pass.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ quality:
 validate: test.requirements test quality
 
 migrate:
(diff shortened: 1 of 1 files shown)


dsjen:
Thanks for the quick fix! 👍

PR #140, opened 2016-09-21
Use date instead of datetime in ModuleEngagementMetricRanges.

dsjen:
* Fixes [AN-7828](https://openedx.atlassian.net/browse/AN-7828).
* Replace datetimes with dates in generated data, unit tests, and model fields for ModuleEngagementMetricRanges.

Both the locally generated data and the unit tests were generating dates as datetimes, and we thus never noticed them locally.  This problem manifests on stage because stage because the pipeline generates the data.  We failed to comply with the contract!


thallada:
Interesting bug! I wonder why this only became a problem after the Django upgrade...

Anyways :+1: glad to get this fixed.

dsjen:
I know!  It's so weird!

PR #142, opened 2016-10-25
AN-7880 Uncomment including "posted_forum" events

thallada:
Now that we have "posted_forum" category events in the course_activity table again, let's re-enable processing those events in the Data API serializer.

Tested locally in analyticstack.

This will pass the data to Insights and we should be able to just flip the switch to get forum events in the graphs and tables.


thallada:
Whoops! I uncommented the test too.

PR #145, opened 2016-12-02
Adds the course metadata enrollment summary endpoint.

dsjen:
* endpoint can be filtered on course IDs and fields
* fake data can be generated for the summary table


dsjen:
Thanks!  Great job to you on the serializer and data generation too!

PR #147, opened 2016-12-06
WIP: Moving to running analytics tests on 16.04

feanil:


feanil:
@thallada @dsjen This moves the tests so that they are run on a 16.04 container instead of 14.04 and replaces coveralls with codecov which is testeng's latest recommendation.

feanil:
@thallada Done. I'll merge once the squashed commit is green unless there is anything else.

PR #149, opened 2016-12-28
Upgraded django, drf, and removed django-model-utils.

dsjen:


clintonb:
FYI @edx/ecommerce

requirements/base.txt
```diff
@@ -1,11 +1,10 @@
 boto==2.42.0                        # MIT
-Django==1.9.9 						# BSD License
+Django==1.10.4 						# BSD License
```
rlucioni: Very nice!

dsjen: :)

PR #150, opened 2017-01-03
AN-7775 Generate local video data

thallada:
Adds new functionality to the `generate_fake_course_data.py` script that will generate fake video data for every video that exists in the course.

This will allow developers to test Insights functionality dealing with videos (e.g. visiting `/engagement/videos/` for a course) without having to point their local Insights server to a real or staging data API.

The authorization set-up is pretty complicated, but this is to avoid having to create a whole OAuth client in the API just for this one developer convenience.

### DIFF
--- a/analytics_data_api/management/commands/generate_fake_course_data.py
+++ b/analytics_data_api/management/commands/generate_fake_course_data.py
@@ -4,12 +4,16 @@ import datetime
 import logging
 import math
 import random
+
 from tqdm import tqdm
 
+from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.utils import timezone
-from analytics_data_api.v0 import models
+
 from analytics_data_api.constants import engagement_events
+from analytics_data_api.v0 import models
+from analyticsdataserver.clients import CourseBlocksApiClient
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
@@ -44,7 +48,7 @@ class Command(BaseCommand):
             '--course_id',
             action='store',
             dest='course_id',
-            default='edX/DemoX/Demo_Course',
+            default='course-v1:edX+DemoX+Demo_Courset',
             help='Course ID for which to generate fake data',
         )
         parser.add_argument(
@@ -175,7 +179,7 @@ class Command(BaseCommand):
 
         logger.info("Generating new weekly course activity data...")
 
-        progress = tqdm(total=math.ceil((end_date - start).days / 7.0) + 1)
+        progress = tqdm(total=math.ceil((end_date - start).days / 7.0))
         while start < end_date:
             active_students = random.randint(100, 4000)
             # End date should occur on Saturday at 23:59:59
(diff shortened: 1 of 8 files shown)


analytics_data_api/management/commands/generate_fake_course_data.py
```diff
@@ -36,7 +38,7 @@ class Command(BaseCommand):
         make_option('-n', '--num-weeks', action='store', type="int", dest='num_weeks',
                     help='Number of weeks worth of data to generate.'),
         make_option('-c', '--course_id', action='store', type='string', dest='course_id',
-                    default='edX/DemoX/Demo_Course', help='Course ID for which to generate fake data'),
+                    default='course-v1:edX+DemoX+Demo_Course', help='Course ID for which to generate fake data'),
```
thallada: @dsjen I'm not sure about this change. I'm pretty confused by edX course IDs in general... do you think most developers (including OpenEdX) would have the new `course-v1` ID over the old `/` delimited one? I had to change it to this to get it to work for me.

analytics_data_api/management/commands/generate_fake_course_data.py
```diff
@@ -161,7 +163,7 @@ def generate_weekly_data(self, course_id, start_date, end_date):
 
         logger.info("Generating new weekly course activity data...")
 
-        progress = tqdm(total=math.ceil((end_date - start).days / 7.0) + 1)
+        progress = tqdm(total=math.ceil((end_date - start).days / 7.0))
```
thallada: This is unrelated to the scope of this PR, but I noticed a off-by-one error in the progress bar here while I was working on this code.

analytics_data_api/management/commands/generate_fake_course_data.py
```diff
@@ -9,8 +9,10 @@
 
 from django.core.management.base import BaseCommand
 from django.utils import timezone
+from django.conf import settings
```
dsjen: Nit: I guess this should go above line 10 to maintain alphabetical ordering. :)

analytics_data_api/management/commands/generate_fake_course_data.py
dsjen: I am for this change!  Thank you!

analyticsdataserver/clients.py
```diff
@@ -0,0 +1,56 @@
...
+                               e.response.status_code)
+            elif e.response.status_code == 404:
+                logger.warning("Course Blocks API failed to return video ids (%s). " +
+                               "Does the course exist in the LMS?",
+                               e.response.status_code)
+            else:
+                logger.warning("Course Blocks API failed to return video ids (%s).", e.response.status_code)
+            return None
+
+        # Setup a terrible hack to silence mysterious flood of ImportErrors from stevedore inside edx-opaque-keys.
```
dsjen: Oh man!  That seems like it was a pain to figure out.

analyticsdataserver/settings/base.py
```diff
@@ -311,6 +311,10 @@
 DEFAULT_PAGE_SIZE = 25
 MAX_PAGE_SIZE = 100
 
+# Only used in a local developer tools script. Override in local.py
+LMS_BASE_URL = 'http://localhost:8000/'
```
dsjen: I'm not sure, but it might make more sense for these to be set to `None` here and then with these default settings in `local.py`.  Just a thought!

thallada: What do you think of removing these from `base.py` altogether and put them in `local.py` with default values and a comment to replace them? These settings should only ever be used in `generate_fake_course_data.py` which is only ever run in the local environment, so I think it makes that more explicit.

analyticsdataserver/clients.py
thallada: Well... I didn't really. I just made the logger shut up :laughing:

analyticsdataserver/settings/base.py
dsjen: ahhh, ok!  I like ti!

PR #152, opened 2017-01-04
Updated the generate fake data parser to use django 1.10.x

dsjen:
I missed this when I upgraded the rest of the API to django 1.10.


dsjen:
If possible, can this be included in the upcoming RC?  It shouldn't affect how the app behaves, but just seems like we should probably "release" the version that works the most. :)

thallada:
Yup, that makes sense. I'll make a new rc.

dsjen:
Thank you!

PR #157, opened 2017-02-08
AN-8236 Collapse "Upcoming" and "Starting soon" availability

thallada:
Insights needs to treat "Starting Soon" the same as "Upcoming".

"Starting Soon" is still stored in the database, but it is collapsed into
the "Upcoming" category in the API view response.

### DIFF
--- a/analytics_data_api/v0/tests/views/test_course_summaries.py
+++ b/analytics_data_api/v0/tests/views/test_course_summaries.py
@@ -22,6 +22,7 @@ class CourseSummariesViewTests(VerifyCourseIdMixin, TestCaseWithAuthentication):
     def setUp(self):
         super(CourseSummariesViewTests, self).setUp()
         self.now = datetime.datetime.utcnow()
+        self.maxDiff = None
 
     def tearDown(self):
         self.model.objects.all().delete()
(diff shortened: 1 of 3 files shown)


analytics_data_api/v0/tests/views/test_course_summaries.py
```diff
@@ -162,3 +162,11 @@ def test_no_matching_courses(self):
     def test_bad_course_id(self, course_ids):
         response = self.authenticated_get(self.path(course_ids=course_ids))
         self.verify_bad_course_id(response)
+
+    def test_collapse_upcoming(self):
+        self.generate_data(availability='Starting Soon')
```
dsjen: I think it would be helpful to generate data for "Upcoming" too and ensure that the result is the sum of the two.

analytics_data_api/v0/tests/views/test_course_summaries.py
```diff
@@ -162,3 +167,15 @@ def test_no_matching_courses(self):
...
+        self.generate_data(course_ids=['foo/bar/baz'], availability='Upcoming')
```
dsjen: I'm not sure that this tests that the summation is happening because the course ID used here is different than the one above.  Maybe I'm missing something?

dsjen: Oh, I'm being silly.  A course would only have one availability type and doesn't need summing.  I'm stick in enrollment summation mode.

PR #166, opened 2017-04-04
AN-8555 Add programs endpoint

thallada:
This PR also adds an option to include a list of program ids that each course is associated with in `course_summaries`. It's disabled by default right now for performance concerns. As we develop the UI in Insights, we can decide whether it is worth enabling.

Because the new `programs` endpoint is so similar to the `course_summaries` endpoint, I refactored the common parts of each into an inheritable abstract view `APIListView` and `APIListViewTestMixin`.

I wanted to try to performance test the new endpoint more, but I'm having trouble doing that right now (I think it might have to do with our VPN being down).

### DIFF
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -555,3 +555,19 @@ class CourseMetaSummaryEnrollmentSerializer(ModelSerializerWithCreatedField, Dyn
         model = models.CourseMetaSummaryEnrollment
         # start_date and end_date used instead of start_time and end_time
         exclude = ('id', 'start_time', 'end_time', 'enrollment_mode')
+
+
+class CourseProgramMetadataSerializer(ModelSerializerWithCreatedField, DynamicFieldsModelSerializer):
+    """
+    Serializer for course and the programs it is under.
+    """
+    course_id = serializers.CharField()
+    program_id = serializers.CharField()
+    program_type = serializers.CharField()
+    program_title = serializers.CharField()
+
+    class Meta(object):
+        model = models.CourseProgramMetadata
+        # excluding course-related fields because the serialized output will be embedded in a course object
+        # with thsoe fields already defined
+        exclude = ('id', 'created')
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -580,17 +580,16 @@ class CourseProgramMetadataSerializer(ModelSerializerWithCreatedField, DynamicFi
     """
     Serializer for course and the programs it is under.
     """
-    course_id = serializers.CharField()
     program_id = serializers.CharField()
     program_type = serializers.CharField()
     program_title = serializers.CharField()
-    courses = serializers.SerializerMethodField()
+    course_ids = serializers.SerializerMethodField()
 
-    def get_courses(self, obj):
-        return obj.get('courses', None)
+    def get_course_ids(self, obj):
+        return obj.get('course_ids', None)
 
     class Meta(object):
         model = models.CourseProgramMetadata
         # excluding course-related fields because the serialized output will be embedded in a course object
-        # with thsoe fields already defined
-        exclude = ('id', 'created')
+        # with those fields already defined
+        exclude = ('id', 'created', 'course_id')
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -511,15 +511,23 @@ class CourseLearnerMetadataSerializer(serializers.Serializer):
 
 class DynamicFieldsModelSerializer(serializers.ModelSerializer):
     """
-    A ModelSerializer that takes an additional `fields` argument that controls which
+    A ModelSerializer that takes additional `fields` and/or `exclude` keyword arguments that control which
     fields should be displayed.
 
     Blatantly taken from http://www.django-rest-framework.org/api-guide/serializers/#dynamically-modifying-fields
+
+    If a field name is specified in both `fields` and `exclude`, then the exclude option takes precedence and the field
+    will not be included in the serialized result.
+
+    Keyword Arguments:
+        fields  -- list of field names on the model to include in the serialized result
+        exclude -- list of field names on the model to exclude in the serialized result
     """
 
     def __init__(self, *args, **kwargs):
         # Don't pass the 'fields' arg up to the superclass
         fields = kwargs.pop('fields', None)
+        exclude = kwargs.pop('exclude', None)
 
         # Instantiate the superclass normally
         super(DynamicFieldsModelSerializer, self).__init__(*args, **kwargs)
@@ -529,6 +537,13 @@ class DynamicFieldsModelSerializer(serializers.ModelSerializer):
             allowed = set(fields)
             existing = set(self.fields.keys())
             for field_name in existing - allowed:
+                self.fields.pop(field_name)
+
+        if exclude is not None:
+            # Drop any fields that are specified in the `exclude` argument.
+            disallowed = set(exclude)
+            existing = set(self.fields.keys())
+            for field_name in existing & disallowed:  # intersection
                 self.fields.pop(field_name)
 
 
@@ -547,9 +562,13 @@ class CourseMetaSummaryEnrollmentSerializer(ModelSerializerWithCreatedField, Dyn
     cumulative_count = serializers.IntegerField(default=0)
     count_change_7_days = serializers.IntegerField(default=0)
     enrollment_modes = serializers.SerializerMethodField()
+    programs = serializers.SerializerMethodField()
 
     def get_enrollment_modes(self, obj):
         return obj.get('enrollment_modes', None)
+
+    def get_programs(self, obj):
+        return obj.get('programs', None)
 
     class Meta(object):
         model = models.CourseMetaSummaryEnrollment
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -584,6 +584,10 @@ class CourseProgramMetadataSerializer(ModelSerializerWithCreatedField, DynamicFi
     program_id = serializers.CharField()
     program_type = serializers.CharField()
     program_title = serializers.CharField()
+    courses = serializers.SerializerMethodField()
+
+    def get_courses(self, obj):
+        return obj.get('courses', None)
 
     class Meta(object):
         model = models.CourseProgramMetadata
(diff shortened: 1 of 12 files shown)


analytics_data_api/v0/views/__init__.py
```diff
@@ -97,3 +108,119 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+        pass  # to be implemented in subclasses
+
+    def default_result(self, item_id):
+        """Default result with fields pre-populated to default values."""
+        result = {
+            self.model_id: item_id,
+        }
+        return result
+
+    def get_result_from_model(self, model, base_result=None, field_list=None):
```
dsjen: I was having trouble understanding this initially and I think it's because it's more than getting the "results," but about updating it.  `update_result_from_model` instead?

analytics_data_api/v0/views/__init__.py
```diff
@@ -97,3 +108,119 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+        fields -- The comma-separated fields to return in the response.
+            For example, 'course_id,created'. Default is to return all fields.
+        exclude -- The comma-separated fields to exclude in the response.
+            For example, 'course_id,created'. Default is to not exclude any fields.
+    """
+    ids = None
+    fields = None
+    exclude = None
+    always_exclude = []
+    serializer_class = serializers.ModelSerializer
```
dsjen: I think setting this here is a bit confusing.  The [GenericAPIView](http://www.django-rest-framework.org/api-guide/generic-views/#genericapiview) docs say that either `serializer_class` or `get_serializer_class()` can be overriden and if downstream we decided for whatever reason to override the method, then our abstractly declared `serializer_class` would be inconsistent.  What do you think about removing this line entirely?

I'm noticing a pattern of setting default values and very interesting in hearing what you think!

analytics_data_api/v0/serializers.py
```diff
@@ -547,11 +562,35 @@ class CourseMetaSummaryEnrollmentSerializer(ModelSerializerWithCreatedField, Dyn
...
+    program_title = serializers.CharField()
+    courses = serializers.SerializerMethodField()
+
+    def get_courses(self, obj):
+        return obj.get('courses', None)
+
+    class Meta(object):
+        model = models.CourseProgramMetadata
+        # excluding course-related fields because the serialized output will be embedded in a course object
+        # with thsoe fields already defined
```
dsjen: ~~thsoe~~ those

analytics_data_api/v0/tests/test_serializers.py
```diff
@@ -0,0 +1,27 @@
...
+from django_dynamic_fixture import G
+
+from analytics_data_api.v0 import models as api_models, serializers as api_serializers
+
+
+class TestSerializer(api_serializers.CourseEnrollmentDailySerializer, api_serializers.DynamicFieldsModelSerializer):
+    pass
+
+
+class DynamicFieldsModelSerializerTests(TestCase):
```
dsjen: Nice test!

analytics_data_api/v0/serializers.py
```diff
@@ -547,11 +562,35 @@ class CourseMetaSummaryEnrollmentSerializer(ModelSerializerWithCreatedField, Dyn
...
         model = models.CourseMetaSummaryEnrollment
         # start_date and end_date used instead of start_time and end_time
         exclude = ('id', 'start_time', 'end_time', 'enrollment_mode')
+
+
+class CourseProgramMetadataSerializer(ModelSerializerWithCreatedField, DynamicFieldsModelSerializer):
+    """
+    Serializer for course and the programs it is under.
+    """
+    course_id = serializers.CharField()
```
dsjen: Can this field be deleted or is it needed?

analytics_data_api/v0/serializers.py
```diff
@@ -547,11 +562,35 @@ class CourseMetaSummaryEnrollmentSerializer(ModelSerializerWithCreatedField, Dyn
...
+    program_id = serializers.CharField()
+    program_type = serializers.CharField()
+    program_title = serializers.CharField()
+    courses = serializers.SerializerMethodField()
```
dsjen: This might be more accurately named "course_ids" instead?

analytics_data_api/v0/views/programs.py
```diff
@@ -0,0 +1,63 @@
...
+        Results can be filed to the course IDs specified or limited to the fields.
+
+        program_ids -- The comma-separated program identifiers for which metadata is requested.
+            Default is to return all programs.
+        fields -- The comma-separated fields to return in the response.
+            For example, 'program_id,created'.  Default is to return all fields.
+        exclude -- The comma-separated fields to exclude in the response.
+            For example, 'course_id,created'.  Default is to not exclude any fields.
+    """
+    always_exclude = ['course_id']  # original model has course_id, but the serializer does not (after aggregation)
```
dsjen: Can we just remove this field from the serializer entirely or do we need it to keep django happy?

analytics_data_api/v0/views/programs.py
```diff
@@ -0,0 +1,63 @@
...
+    **Parameters**
+
+        Results can be filed to the course IDs specified or limited to the fields.
+
+        program_ids -- The comma-separated program identifiers for which metadata is requested.
+            Default is to return all programs.
+        fields -- The comma-separated fields to return in the response.
+            For example, 'program_id,created'.  Default is to return all fields.
+        exclude -- The comma-separated fields to exclude in the response.
+            For example, 'course_id,created'.  Default is to not exclude any fields.
```
dsjen: ~~course_id~~ program_id

analytics_data_api/v0/views/course_summaries.py
```diff
@@ -37,101 +33,100 @@ class CourseSummariesView(generics.ListAPIView):
...
-        if self.course_ids is not None:
-            for course_id in self.course_ids:
-                validate_course_id(course_id)
-
-        return super(CourseSummariesView, self).get(request, *args, **kwargs)
-
-    def default_summary(self, course_id, count_fields):
+        programs = split_query_argument(query_params.get('programs'))
+        if not programs:
+            self.always_exclude = ['programs']
```
dsjen: It might be a little safer to append "programs" in case `always_exclude` was set.

analytics_data_api/v0/views/programs.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    always_exclude = ['course_id']  # original model has course_id, but the serializer does not (after aggregation)
+    serializer_class = serializers.CourseProgramMetadataSerializer
+    model = models.CourseProgramMetadata
+    model_id = 'program_id'
+    program_meta_fields = ['program_type', 'program_title']
+
+    def default_result(self, item_id):
+        """Default program with id, empty metadata, and empty courses array."""
+        program = {
```
dsjen: This could call the super method?

analytics_data_api/v0/views/programs.py
thallada: Thanks for pointing that out. I just tried out the serializer without the `course_id` field and it works fine!

analytics_data_api/v0/views/programs.py
thallada: Yes, it was supposed to!

analytics_data_api/v0/views/__init__.py
thallada: Sounds good to me. I'll also remove the `model` assignment here too.

analytics_data_api/v0/views/__init__.py
```diff
@@ -97,3 +108,119 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+    model = models.Model
+    model_id = 'id'
```
thallada: :+1: naming is hard, definitely keep letting me know if you have better variable name suggestions.

analytics_data_api/v0/views/__init__.py
```diff
@@ -97,3 +108,119 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+        self.exclude = self.always_exclude + (exclude if exclude else [])
+        self.ids = split_query_argument(query_params.get('ids'))
+        self.verify_ids()
+
+        return super(APIListView, self).get(request, *args, **kwargs)
+
+    def verify_ids(self):
+        pass  # to be implemented in subclasses
+
+    def default_result(self, item_id):
```
thallada: The result of turning the model object instance into a python dictionary, and one step before serializing it into a json response. Not sure what to call that... maybe field_dict (`base_field_dict`, `update_field_dict_from_model`, `postprocess_field_dict`)?

analytics_data_api/v0/views/__init__.py
```diff
@@ -97,3 +108,119 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+
+            for model in model_group:
+                result = self.get_result_from_model(model, base_result=result)
+
+            result = self.postprocess_result(result)
+            formatted_results.append(result)
+
+        return formatted_results
+
+    def get_query(self):
```
thallada: I did try that double star, but the Q function didn't seem to work with that unfortunately.

analytics_data_api/v0/views/__init__.py
```diff
@@ -97,3 +108,119 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+        return result
+
+    def get_result_from_model(self, model, base_result=None, field_list=None):
+        field_list = (field_list if field_list else
+                      [f.name for f in self.model._meta.get_fields()])  # pylint: disable=protected-access
+        result = base_result if base_result else {}
+        result.update({field: getattr(model, field) for field in field_list})
+        return result
+
+    def postprocess_result(self, result):
```
thallada: Right, this is called after a single python dictionary has been created from a group of models (grouped by `course_id` in the case of `course_summaries` and `program_id` in `programs`).

So this is just tacking any extra fields on that don't come from the models. `course_summaries` uses it to merge professional and non-verified professional, collapse availability, and add programs_id array.

analytics_data_api/v0/views/__init__.py
dsjen: Ok, thanks!

analytics_data_api/v0/views/__init__.py
dsjen: Naming is the worst and so subjective.  Thanks for considering the suggestion!

analytics_data_api/v0/views/__init__.py
dsjen: Yeah!  I think `field_dict` is a bit clearer too.

analytics_data_api/v0/views/__init__.py
dsjen: Bummer!  Thanks for trying. :(

thallada:
@dsjen I think I've addressed your comments if you would like to take a second look.

PR #170, opened 2017-05-22
Add passing_users to course summary

efischer19:
# [EDU-153](https://openedx.atlassian.net/browse/EDUCATOR-153)

This feels too easy - am I missing something?

### DIFF
--- a/analytics_data_api/management/commands/generate_fake_course_data.py
+++ b/analytics_data_api/management/commands/generate_fake_course_data.py
@@ -161,15 +161,17 @@ class Command(BaseCommand):
             progress.update(1)
             date = date + datetime.timedelta(days=1)
 
-        for mode, ratio in enrollment_mode_ratios.iteritems():
+        for index, (mode, ratio) in enumerate(enrollment_mode_ratios.iteritems()):
             count = int(ratio * daily_total)
+            pass_rate = min(random.normalvariate(.45 + (.1 * index), .15), 1.0)
             cumulative_count = count + random.randint(0, 100)
             models.CourseMetaSummaryEnrollment.objects.create(
                 course_id=course_id, catalog_course_title='Demo Course', catalog_course='Demo_Course',
                 start_time=timezone.now() - datetime.timedelta(weeks=6),
                 end_time=timezone.now() + datetime.timedelta(weeks=10),
                 pacing_type='self_paced', availability='Starting Soon', enrollment_mode=mode, count=count,
-                cumulative_count=cumulative_count, count_change_7_days=random.randint(-50, 50))
+                cumulative_count=cumulative_count, count_change_7_days=random.randint(-50, 50),
+                passing_users=int(cumulative_count * pass_rate))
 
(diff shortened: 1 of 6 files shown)


dsjen:
I just realized that we probably want to ensure that the `exclude` field can be used to exclude the `passing_users` field from the payload in both the top level and in the enrollment modes, since this functionality will be turned on later.  Please let me know if that doesn't make sense!

efischer19:
No problem, I'll get that added this afternoon

analytics_data_api/management/commands/generate_fake_course_data.py
```diff
@@ -161,15 +161,17 @@ def generate_daily_data(self, course_id, start_date, end_date):
             progress.update(1)
             date = date + datetime.timedelta(days=1)
 
-        for mode, ratio in enrollment_mode_ratios.iteritems():
+        for index, (mode, ratio) in enumerate(enrollment_mode_ratios.iteritems()):
             count = int(ratio * daily_total)
+            pass_rate = min(random.normalvariate(.45 + (.1 * index), .15), 1.0)
```
efischer19: I utilize `index` here so we can get a higher pass rate for `verified` than we have for `audit` (scales with position in `enrollment_mode_ratios); is that overly complex for this data?

I also just guessed at the baseline pass rates (from 45% for audit up to 85% for verified), LMK if there's a more accurate value we ought to use.

analytics_data_api/v0/tests/views/test_course_summaries.py
```diff
@@ -159,3 +163,10 @@ def test_programs(self):
         response = self.authenticated_get(self.path(exclude=self.always_exclude[:1], programs=['True']))
         self.assertEquals(response.status_code, 200)
         self.assertItemsEqual(response.data, self.all_expected_results(programs=True))
+
+    @ddt.data('passing_users', )
+    def test_exclude(self, field):
+        self.generate_data()
+        response = self.authenticated_get(self.path(exclude=[field]))
+        self.assertEquals(response.status_code, 200)
+        self.assertEquals(str(response.data).count(field), 0)
```
efischer19: `self.assertNotIn(field, response.data)` only works for the top-level (cumulative) field, not the per-mode field. This ensures the field is not present anywhere.

analytics_data_api/management/commands/generate_fake_course_data.py
dsjen: ❤️

analytics_data_api/v0/tests/views/test_course_summaries.py
dsjen: Nice!

analytics_data_api/v0/views/course_summaries.py
```diff
@@ -120,6 +121,10 @@ def postprocess_field_dict(self, field_dict):
             # don't do expensive looping for programs if we are just going to throw it away
             field_dict = self.add_programs(field_dict)
 
+        for field in self.exclude:
```
dsjen: Thanks for taking care of that!

PR #171, opened 2017-06-07
Travis, why do you do the things you do?

efischer19:
Trying to figure out why https://travis-ci.org/edx/edx-analytics-data-api/builds/240498425 failed


efischer19:
The `requests` package has had a *lot* of releases recently, I'm gonna go ahead and say that's why these builds have been failing (previous success at https://travis-ci.org/edx/edx-analytics-data-api/builds were before requests cut their 2.15.0 tag). See "Release History" at https://pypi.python.org/pypi/requests

Pinning `urllib3` seems to fix things, should we just merge that @edx/educator-dahlia? I'm open to moving the requirement spec into another file under `requirements/`, I'll go look up the license details now.

PR #172, opened 2017-06-09
Patch bump to django.

dsjen:
Instead of going to 1.11.1 when this was originally upgraded, I left this at 1.11 because of a GDAL detection error that caused the build to fail.  This is now resolved in [1.11.2](https://docs.djangoproject.com/en/1.11/releases/1.11.2/).

I was going to make a ticket to upgrade, but this seemed faster?

Please review, @edx/educator-dahlia


dsjen:
@efischer19 -- this PR is so principally agile!

PR #173, opened 2017-06-12
EDUCATOR-464: Add POST method to course_summaries/

kdmccormick:
The course_summaries/ currently supports only the GET method. A client can restrict the number of course summaries returned from the endpoint by passing in a list of desired course IDs in the `course_ids` parameter. However, this is not practically usable, as the number of IDs many clients would want to pass in would make the URL too long. So, Insights currently does not use the `course_ids` parameter, and instead fetches *every* course summary and filters them on the client side. Fetching every summary incurs a response time of ~5s.

To fix this, this PR adds a POST method to this endpoint. The POST method functions exactly the same as the GET method, except that all parameters are passed in the request body, allowing a much larger number of course IDs to be passed in. (It is understood that this is not the semantic meaning of the POST verb, however, this is not an uncommon workaround to URL length restrictions).

[This PR](https://github.com/edx/edx-analytics-data-api-client/pull/33/commits) updates the Analytics API Client to use the new POST method, and [this PR](https://github.com/edx/edx-analytics-dashboard/pull/690) updates Insights to take advantage of the course ID filtering now usable in the client.

JIRA Ticket: [EDUCATOR-464](https://openedx.atlassian.net/browse/EDUCATOR-464).

TODO:
- [x] Write tests
- [x] Get tests to pass
- [ ] Get thumbs

@edx/educator-dahlia

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -9,3 +9,4 @@ Jason Bau <jbau@stanford.edu>
 John Jarvis <jarv@edx.org>
 Dmitry Viskov <dmitry.viskov@webenterprise.ru>
 Eric Fischer <efischer@edx.org>
+Kyle McCormick <kylemccor@gmail.com>
--- a/analytics_data_api/v0/tests/views/__init__.py
+++ b/analytics_data_api/v0/tests/views/__init__.py
@@ -100,14 +100,28 @@ class APIListViewTestMixin(object):
     list_name = 'list'
     default_ids = []
     always_exclude = ['created']
+    test_post_method = False
 
-    def path(self, ids=None, fields=None, exclude=None, **kwargs):
-        query_params = {}
-        for query_arg, data in zip([self.ids_param, 'fields', 'exclude'], [ids, fields, exclude]) + kwargs.items():
-            if data:
-                query_params[query_arg] = ','.join(data)
-        query_string = '?{}'.format(urlencode(query_params))
+    def path(self, query_data=None):
+        query_data = query_data or {}
+        concat_query_data = {param: ','.join(arg) for param, arg in query_data.items() if arg}
+        query_string = '?{}'.format(urlencode(concat_query_data)) if concat_query_data else ''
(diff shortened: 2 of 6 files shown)


efischer19:
The failure you're currently seeing will go away with a `git pull; git rebase origin/master` to pick up the changes in https://github.com/edx/edx-analytics-data-api/pull/171

analytics_data_api/v0/tests/views/__init__.py
```diff
@@ -100,15 +100,28 @@ class APIListViewTestMixin(object):
...
-        for query_arg, data in zip([self.ids_param, 'fields', 'exclude'], [ids, fields, exclude]) + kwargs.items():
-            if data:
-                query_params[query_arg] = ','.join(data)
-        query_string = '?{}'.format(urlencode(query_params))
+    def path(self, query_data={}):
+        concat_query_data = {param: ','.join(arg) for param, arg in query_data.items() if arg}
+        query_string = '?{}'.format(urlencode(concat_query_data)) if concat_query_data else ''
         return '/api/v0/{}/{}'.format(self.list_name, query_string)
 
+    def validated_request(self, ids=None, fields=None, exclude=None, **extra_args):
```
kdmccormick: This is the function I'm most interested in feedback on. I'm making it so every request is done as both a GET and POST, and the results are compared before returning. Let me know what you think @efischer19.

efischer19: I like it, very nice test addition

AUTHORS
```diff
@@ -9,3 +9,4 @@ Jason Bau <jbau@stanford.edu>
 John Jarvis <jarv@edx.org>
 Dmitry Viskov <dmitry.viskov@webenterprise.ru>
 Eric Fischer <efischer@edx.org>
+Kyle McCormick <kylemccor@gmail.com>
```
dsjen: :1st_place_medal:

PR #184, opened 2018-02-23
WIP - Support for enterprise data api

georgebabey:
TODO:
- [ ] Dynamically pull in enterprise app
- [ ] Dynamically pull in router config for enterprise
- [ ] Upgrade edx-drf-extensions lib version once built

### DIFF
--- a/analyticsdataserver/settings/base.py
+++ b/analyticsdataserver/settings/base.py
@@ -205,8 +205,13 @@ LOCAL_APPS = (
     'analytics_data_api.v0',
 )
 
+# TODO: Make optional
+OPTIONAL_APPS = (
+    'enterprise_data',
+)
+
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#installed-apps
-INSTALLED_APPS = DJANGO_APPS + THIRD_PARTY_APPS + LOCAL_APPS
+INSTALLED_APPS = DJANGO_APPS + THIRD_PARTY_APPS + LOCAL_APPS + OPTIONAL_APPS
 ########## END APP CONFIGURATION
 
 
@@ -242,6 +247,11 @@ LOGGING = {
             'level': 'DEBUG',
             'propagate': True
         },
+        'enterprise_data': {
+            'handlers': ['console'],
+            'propagate': True,
+            'level': 'INFO'
+        },
         # See https://elasticutils.readthedocs.io/en/latest/debugging.html
         # INFO-level logs tell us when nodes fail and are resurrected.
         'elasticsearch': {
@@ -256,7 +266,7 @@ LOGGING = {
         'elasticsearch.trace': {
             'handlers': ['null'],
             'propagate': False
-        }
+        },
     },
 }
 ########## END LOGGING CONFIGURATION
@@ -318,6 +328,14 @@ AGGREGATE_PAGE_SIZE = 10
 # We need to set this to None so that we can pass in a large number of Course IDs
 # to course_summaries/
 DATA_UPLOAD_MAX_NUMBER_FIELDS = None
+
+JWT_AUTH = {
+    'JWT_ALGORITHM': 'HS256',
+    'JWT_AUDIENCE': 'lms-key',
+    'JWT_ISSUER': 'http://127.0.0.1:8000/oauth2',
+    'JWT_DECODE_HANDLER': 'edx_rest_framework_extensions.utils.jwt_decode_handler',
+    'JWT_VERIFY_AUDIENCE': False,
+}
 
 ########## END ANALYTICS DATA API CONFIGURATION
 
--- a/analyticsdataserver/settings/local.py
+++ b/analyticsdataserver/settings/local.py
@@ -62,9 +62,11 @@ SWAGGER_SETTINGS = {
     'api_key': 'edx'
 }
 
+ENABLE_ENTERPRISE_INTEGRATION = True
+
 # These two settings are used in generate_fake_course_data.py.
 # Replace with correct values to generate local fake video data.
-LMS_BASE_URL = 'http://localhost:8000/'  # the base URL for your running local LMS instance
+LMS_BASE_URL = 'http://localhost:18000/'  # the base URL for your running local LMS instance
 COURSE_BLOCK_API_AUTH_TOKEN = 'paste auth token here'  # see README for instructions on how to configure this value
 
 # In Insights, we run this API as a separate service called "analyticsapi" to run acceptance/integration tests. Docker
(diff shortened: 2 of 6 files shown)


analyticsdataserver/settings/base.py
```diff
@@ -205,8 +205,12 @@
     'analytics_data_api.v0',
 )
 
+OPTIONAL_APPS = (
+    'enterprise_data',
+)
```
douglashall: Seems like this should be in `THIRD_PARTY_APPS`, there really isn't anything optional about it.

analyticsdataserver/settings/base.py
```diff
@@ -293,6 +302,7 @@
 ########## ANALYTICS DATA API CONFIGURATION
 
 ANALYTICS_DATABASE = 'default'
+APPS_LABELS_FOR_ANALYTICS_DATABASE = ('v0', 'enterprise_data')
```
douglashall: What's going on here? The `app_label` for the existing API is `v0`? That's a bummer.

I wouldn't make this a setting. Just make a list [here](https://github.com/edx/edx-analytics-data-api/pull/184/files#diff-110b488f2b07b14fdb4a2589e7da9e18R10).

analyticsdataserver/settings/local.py
```diff
@@ -62,6 +62,8 @@
     'api_key': 'edx'
 }
 
+ENABLE_ENTERPRISE_INTEGRATION = True
```
douglashall: I question the value of this setting. I could see an argument being made for having a waffle switch that would allow you to turn the enterprise-related endpoints on/off at runtime. Doing it at deployment doesn't seem all that useful.

analyticsdataserver/settings/base.py
```diff
@@ -242,6 +246,11 @@
             'level': 'DEBUG',
             'propagate': True
         },
+        'enterprise_data': {
+            'handlers': ['console'],
+            'propagate': True,
+            'level': 'INFO'
```
douglashall: Nit: Swap order of level/propagate to make it the same as the other loggers.

analyticsdataserver/settings/local.py
```diff
@@ -74,4 +76,11 @@
 # hosts.
 ALLOWED_HOSTS = ['localhost', '127.0.0.1', '::1', 'analyticsapi']
 
+JWT_AUTH.update({
```
brittneyexline: out of curiosity, how does this get updated for other environments like prod?

requirements/base.txt
```diff
@@ -17,3 +17,4 @@ edx-ccx-keys==0.2.1
 edx-django-release-util==0.3.0
 edx-opaque-keys==0.4.0
 edx-rest-api-client==1.4.0          # Apache 2.0
+edx-drf-extensions==1.2.4
```
brittneyexline: does edx-enterprise-data need to be pulled in here too? (i'm guessing it's not on pypi yet...)

analyticsdataserver/settings/local.py
georgebabey: This gets set in the configuration repo: https://github.com/edx/configuration/blob/f3d6a97a405321cc54f86727755a2d5f33efb5a1/playbooks/roles/edxapp/defaults/main.yml#L1104

analyticsdataserver/urls.py
```diff
@@ -24,5 +24,8 @@
     admin.autodiscover()
     urlpatterns.append(url(r'^site/admin/', include(admin.site.urls)))
 
+if  'enterprise_data' in settings.INSTALLED_APPS:
```
douglashall: But it will always be in `INSTALLED_APPS` according to your base settings file. If installing the `edx-enterprise-data` app needs to be optional, we should probably come up with a real way of doing that. Isn't this the problem that Nimisha solved for edx-platform?

douglashall: @nasthagiri ^^

georgebabey: @douglashall Yes, the plugin architecture would allow this, but it is also currently coupled to edx-platform.

My thought here was that one could remove enterprise_data from the THIRD_PARTY_APPS and not have to update a bunch of other places such as this `urls.py` file

douglashall: But wouldn't you want them to remove this? Otherwise you end up with dead code.

georgebabey: I see it as a configuration switch, just the configuration happens to be if the enterprise_data app is installed. If its not installed, this code will get skipped.

nasthagiri: FWIW, the newly created plugin architecture in edx-platform can be easily extracted into its own repo.  The fundamental design and implementation are generic so can apply to any other IDA.
There are a few Enums that specifically mention LMS and Studio - but those Enums can remain in edx-platform.  Let me know if you need further input if your team chooses to take on the extraction. Thanks.

requirements/production.txt
```diff
@@ -4,5 +4,6 @@
 
 MySQL-python==1.2.5	# GPL License
 PyYAML==3.11		# MIT
+gevent==1.0.2
```
georgebabey: This was required for the ansible deploys when we moved to edx_django_service role, we're now using gevent worker w/ gunicorn

georgebabey:
@douglashall @brittneyexline Can you give this another look?

georgebabey:
@thallada Correct, I didn't change the default authentication, only the new enterprise endpoint uses JWT auth here: https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/api/v0/views.py#L23

analyticsdataserver/settings/base.py
```diff
@@ -297,6 +303,8 @@
 
 ENABLE_ADMIN_SITE = False
 
+LMS_BASE_URL = ''
```
dsjen: Nit: this setting isn't used too much, but is referenced https://github.com/edx/edx-analytics-data-api/blob/5c635010f15650670a46af1ca52c14ad706d8f8a/analytics_data_api/management/commands/generate_fake_course_data.py#L300. It makes sense to have a default to this, but perhaps `None` instead?

georgebabey: This was added to base because the enterprise-data api endpoint needs it as well (for now, someday we'd like to get the enterprise association into the JWT, but for now we are making a call to LMS to determine)

dsjen: Gotcha! It totally makes sense to have that here. I'm assuming it would be actually set in the configuration repo.

Does it mess up the enterprise-data api endpoint by setting it to `None` in `base.py`? My one concern is that if this isn't set to `None`, the `AttributeError` in https://github.com/edx/edx-analytics-data-api/blob/5c635010f15650670a46af1ca52c14ad706d8f8a/analytics_data_api/management/commands/generate_fake_course_data.py#L302 won't be hit for other environments. We could certainly update that piece of code to check for an empty string too...

georgebabey: Oh good catch! Will update it to `None`

dsjen: 🍰

PR #185, opened 2018-04-12
Revert "Support for enterprise-data-api endpoint"

georgebabey:
This reverts commit 142de26e2c8c1038275d380608433d928bc379d4.

Reverting this until the configuration changes are approved in https://github.com/edx/configuration/pull/4443 as it is breaking sandbox builds


georgebabey:
Need settings.LMS_BASE_URL set in configuration otherwise the enterprise-data app fails to load:

API_BASE_URL = settings.LMS_BASE_URL + 'enterprise/api/v1/'\nTypeError: unsupported operand type(s) for +: 'NoneType' and 'unicode'",

PR #194, opened 2018-06-07
Play Nice with LMS Docker

BessieSteinberg:
- This was pointing to the wrong port for LMS docker, now fix


dsjen:
@BessieSteinberg -- I think you checked in elastic search?

elasticsearch-1.5.2/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
```
thallada: I don't think we want to include these files in the repo. Maybe it would be good to add `elasticsearch-1.5.2/` and `elasticsearch-1.5.2.zip` to the `.gitignore`.

BessieSteinberg:
@dsjen oops! you are right! i'll take those files out

PR #199, opened 2018-08-01
Allow CORS for API access for local development

georgebabey:
This enables CORS during local development. localhost:1991 is whitelisted
which is the default port used by standalone frontends built from the
edx-frontend-cookiecutter.

This is necessary for the edx-portal to connect to the data-api during local development. Because we're using authorization header and not the session cookie we didn't need `CORS_ALLOW_CREDENTIALS`


georgebabey:
Because we're using authorization header and not the session cookie we didn't need `CORS_ALLOW_CREDENTIALS`, otherwise this matches other IDAs.

@tuchfarber You were right, we needed this. This is equivalent to your journals change here: https://github.com/edx/journals/commit/4cf20a71408f8b8fd7e0826b62bce0b33c459aba#diff-7fe11226cff646f5d9f35faa76217059

PR #204, opened 2018-08-24
Updating README with some updated information for loading certain bit…

christopappas:
…s of data for dev setup


georgebabey:
@christopappas because this is specific to edx-enterprise-data I think it is most logical to have the README there updated.

PR #211, opened 2018-09-24
edx-enterprise-data version bump

mushtaqak:


AUTHORS
```diff
@@ -12,3 +12,4 @@ Eric Fischer <efischer@edx.org>
 Kyle McCormick <kylemccor@gmail.com>
 George Babey <gbabey@edx.org>
 Noraiz Anwar <noraizbhatti@gmail.com>
+Mushtaq Ali <mushtaak@gmail.com>
```
asadiqbal08: can you add me too here ?

mushtaqak: its easy

asadiqbal08: ok it is different repo.

PR #221, opened 2018-10-23
Add JWT_AUTH_COOKIE setting to enable JwtAuthentication on DRF API endpoints.

douglashall:


analyticsdataserver/settings/base.py
```diff
@@ -338,6 +338,7 @@
 JWT_AUTH = {
     'JWT_ALGORITHM': 'HS256',
     'JWT_AUDIENCE': 'lms-key',
+    'JWT_AUTH_COOKIE': 'edx-jwt-cookie',
```
robrap: @douglashall: I thought this was meant to have a default.  Is there a problem?
See https://github.com/edx/edx-drf-extensions/blob/999efe54be2f2a4c115deda1fc2b88df339ad8ab/edx_rest_framework_extensions/auth/jwt/cookies.py#L9

douglashall: Yeah, the problem is that if JWT_AUTH_COOKIE is not set, the [django-rest-framework-jwt](http://getblimp.github.io/django-rest-framework-jwt/#additional-settings) code will not check the cookie for the JWT.

robrap: I see. I think that means that default code is misleading and should be removed and replaced with a comment regarding why a default is not possible (I.e. that django-rest-framework-jwt expects the setting). Does this make sense to you?

douglashall: Yes, I see no reason for the default. @nasthagiri Is the default JWT_AUTH_COOKIE value referenced above useful in any way?

nasthagiri: +1 on getting rid of the default.

PR #226, opened 2018-10-26
Bumping edx-enterprise-data to 1.0.6

christopappas:


christopappas:
Noted! Tests look good locally, so I'll wait for these to pass then ship to stage~

PR #237, opened 2018-12-07
bjh/recent count change

inventhouse:
Added `recent_count_change` to `course_summaries` endpoint for DISCO-183; I have not yet surfaced this in the analytics-data-api-client

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ clean:
 	coverage erase
 
 test: clean
+	if [ -e elasticsearch-$(ELASTICSEARCH_VERSION) ]; then curl --silent --head http://localhost:$(ELASTICSEARCH_PORT)/roster_test > /dev/null || make test.run_elasticsearch; fi  # Launch ES if installed and not running
 	coverage run ./manage.py test --settings=$(TEST_SETTINGS) \
 		--with-ignore-docstrings --exclude-dir=analyticsdataserver/settings \
 		$(PACKAGES)
--- a/analytics_data_api/v0/tests/views/__init__.py
+++ b/analytics_data_api/v0/tests/views/__init__.py
@@ -104,7 +104,10 @@ class APIListViewTestMixin(object):
 
     def path(self, query_data=None):
         query_data = query_data or {}
-        concat_query_data = {param: ','.join(arg) for param, arg in query_data.items() if arg}
+        concat_query_data = {
+            param: arg if isinstance(arg, str) else ','.join(arg)
+            for param, arg in query_data.items() if arg
+        }
         query_string = '?{}'.format(urlencode(concat_query_data)) if concat_query_data else ''
         return '/api/v0/{}/{}'.format(self.list_name, query_string)
 
--- a/analytics_data_api/v0/tests/views/test_course_summaries.py
+++ b/analytics_data_api/v0/tests/views/test_course_summaries.py
@@ -33,7 +33,8 @@ class CourseSummariesViewTests(VerifyCourseIdMixin, TestCaseWithAuthentication, 
         self.model.objects.all().delete()
 
     def create_model(self, model_id, **kwargs):
-        for mode in kwargs['modes']:
+        modes = kwargs.get('modes', [])
+        for mode in modes:
             G(self.model, course_id=model_id, catalog_course_title='Title', catalog_course='Catalog',
               start_time=datetime.datetime(2016, 10, 11, tzinfo=pytz.utc),
               end_time=datetime.datetime(2016, 12, 18, tzinfo=pytz.utc),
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -49,6 +49,7 @@ clean:
 	coverage erase
 
 test: clean
+	curl --silent --head http://localhost:9223/roster_test > /dev/null || make test.run_elasticsearch  # Launch ES if needed
```
inventhouse: This is really helpful locally, but fails in Travis; if someone knows how to make it work there, that'd be great, otherwise I'll pull it before merging

christopappas: it looks like the error in travis is saying there's a directory missing, a directory that is created locally by running the `make test.install_elasticseach` command (https://github.com/edx/edx-analytics-data-api/blob/9190ca68e2bf9033b019dfd9a6f7c7ba84d4ff0a/Makefile#L19)

i wonder if it makes sense to have `test.run_elasticsearch` run the `test.install_elasticsearch` logic if the folder is not found?

that said though, because we haven't hit this problem in the past, running ES in travis may not be necessary (specifically if we were not running it before), and would just be adding this to travis only to make local dev a little more convenient

what do you think?

inventhouse: If there's an easy way to command to check for "am I on Travis?" I could probably add that in to avoid the launch attempt; do you know of something I could check for?

inventhouse: Maybe just "if that directory doesn't exist, don't try" (since it would fail locally too)

dianakhuang: I haven't been following closely, but Travis has environment variables it includes that you can check for: https://docs.travis-ci.com/user/environment-variables#default-environment-variables

inventhouse: I added a check for the local ES install and only attempt to launch from there if it exists.  That passes Travis.

christopappas: Cool to see you got that working! Nice!

inventhouse:
Added unit tests and squashed, requesting re-review

analytics_data_api/v0/tests/views/test_course_summaries.py
```diff
@@ -33,7 +33,8 @@ def tearDown(self):
         self.model.objects.all().delete()
 
     def create_model(self, model_id, **kwargs):
-        for mode in kwargs['modes']:
+        modes = kwargs['modes']
```
christopappas: Could the modes key be missing from kwargs? Would `kwargs.get('modes', [])` be safer?

analytics_data_api/v0/tests/views/test_course_summaries.py
```diff
@@ -103,13 +108,20 @@ def expected_result(self, item_id, modes=None, availability='Current', programs=
...
+                             availability='Current',
+                             programs=False,
+                             recent_count_change=None):
         if modes is None:
             modes = enrollment_modes.ALL
 
-        return super(CourseSummariesViewTests, self).all_expected_results(ids=ids, modes=modes,
-                                                                          availability=availability,
-                                                                          programs=programs)
+        return super(CourseSummariesViewTests, self)\
```
christopappas: I don't like backslashes, so this is just a personal preference recommendation, but, how do you think this looks:

```
        return super(CourseSummariesViewTests, self).all_expected_results(
            ids=ids,
            modes=modes,
            availability=availability,
            programs=programs,
            recent_count_change=recent_count_change
        )
```

analytics_data_api/v0/views/course_summaries.py
```diff
@@ -154,12 +182,27 @@ def postprocess_field_dict(self, field_dict):
...
         for field in self.exclude:
             for mode in field_dict['enrollment_modes']:
                 _ = field_dict['enrollment_modes'][mode].pop(field, None)
 
         return field_dict
 
+    def add_recent_count_change(self, field_dict):
+        # Pick the most recent daily count on-or-before the desired date
+        # If there are no daily counts before the desired date, assume 0
+        recents = models.CourseEnrollmentDaily.objects\
```
christopappas: It's cool if the linter if happy, but how I would probably tweak:

```
            recents = models.CourseEnrollmentDaily.objects.filter(
                course_id=field_dict['course_id'],
                date__lte=self.recent_date
            ).order_by('-date')[0:1]
```

analytics_data_api/v0/tests/views/__init__.py
```diff
@@ -104,7 +104,11 @@ class APIListViewTestMixin(object):
 
     def path(self, query_data=None):
         query_data = query_data or {}
-        concat_query_data = {param: ','.join(arg) for param, arg in query_data.items() if arg}
+        concat_query_data = \
```
christopappas: Personal style preference

```
            concat_query_data = {
                param: arg if isinstance(arg, str) else ','.join(arg)
                for param, arg in query_data.items() if arg
            }
```

analytics_data_api/v0/tests/views/test_course_summaries.py
inventhouse: That would be safer and I'll change it, though the previous implementation would have errored if 'modes' was missing

analytics_data_api/v0/tests/views/test_course_summaries.py
inventhouse: I like that better, I'll see what pylint thinks of that, butit didn't like the variant of this that I tried.  There's a lot of things I disagree with pylint/pep8 about.

analytics_data_api/v0/views/course_summaries.py
inventhouse: I'll try it, I agree it's more readable

analytics_data_api/v0/tests/views/__init__.py
inventhouse: Same

inventhouse:
All the backslash nonsense were just to appease pylint; these suggestions all worked and are much more readable.  Assuming these changes pass, I'll merge them.

PR #254, opened 2019-03-19
edx-enterprise-data version bump to 1.0.18

muhammad-ammar:
[ENT-1602](https://openedx.atlassian.net/browse/ENT-1602)

Updating `edx-enterprise-data` version to incorporate latest changes from [1.0.18](https://github.com/edx/edx-enterprise-data/releases/tag/1.0.18)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,3 +15,4 @@ Noraiz Anwar <noraizbhatti@gmail.com>
 Mushtaq Ali <mushtaak@gmail.com>
 Zubair Afzal <zubair.fast@gmail.com>
 Saleem Latif <saleem_ee@hotmail.com>
+Muhammad Ammar <mammar@gmail.com>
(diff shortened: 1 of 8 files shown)


muhammad-ammar:
@brittneyexline `make upgrade` is updating version of lots of existing pkgs and I am not sure if updated pkgs are good to go? Also I am not sure when to run `make upgrade` and when not because I saw PRs where people just updated a pkg version just in `.txt` files without doing `make upgrade`.

brittneyexline:
@muhammad-ammar i would say the good citizen thing to do is to go forward with getting all requirements up to the latest versions, even if they are unrelated, and generally if tests still pass, it's probably fine. that said, it can be bumpy sometimes. i will give it a shot myself, but if you don't see updates on this PR before you get back online, i would advise at the very least making sure django rules and edx-rbac are in the dependencies, otherwise this will probably not behave as expected when deployed.

muhammad-ammar:
@brittneyexline Executed `make upgrade` and now all the `.txt` requirement files are updated correctly.

PR #255, opened 2019-03-21
ENT-1718:  Django Admin updates for user role based permissions

muhammad-ammar:
__Description:__

This PR removes the `ENABLE_ADMIN_SITE` so that admin is enabled by default. It also adds the new `enterprise_data_roles` app into `INSTALLED_APPS`.

**Testing Instructions:**

1. Create a superuser for django admin
2. Run `python manage.py runserver`
2. Goto http://127.0.0.1:8000/admin
3. Open `Enterprise data role assignments` model
4. Create a role assignment and it should save successfully
5. Open `Switches` model under `WAFFLE` app
6. Create/Update/Delete switch should work as expected


analyticsdataserver/settings/base.py
```diff
@@ -315,8 +316,6 @@
 ANALYTICS_DATABASE = 'default'
 DATABASE_ROUTERS = ['analyticsdataserver.router.AnalyticsApiRouter']
 
-ENABLE_ADMIN_SITE = False
```
muhammad-ammar: @mulby Any idea about why admin site is disabled in production? We need to enable this now and wanted to have your opinion.

mulby: I don't remember, sorry. My guess is that we felt like it wasn't worth the security risk if we didn't have a need for it.

PR #266, opened 2019-05-20
update edx-enterprise-data version

muhammad-ammar:


analyticsdataserver/settings/local.py
```diff
@@ -89,7 +89,7 @@
 })
 
 CORS_ORIGIN_WHITELIST = (
-    'localhost:1991'
+    'http://localhost:1991',
```
muhammad-ammar: **Note to Reviewers:** Latest changes in `django-cors-headers` requires scheme in origins. Please [see](https://github.com/ottoyiu/django-cors-headers/pull/397)

PR #267, opened 2019-05-30
Move analytics defaults from configration repo

syedimranhassan:


analyticsdataserver/settings/base.py
```diff
@@ -376,3 +396,5 @@
 }
 
 ########## EDX ENTERPRISE DATA CONFIGURATION
+API_AUTH_TOKEN = 'put-your-api-token-here'
+CSRF_COOKIE_SECURE = False
```
coryleeio: You need to include all these values as well:
https://github.com/edx/configuration/blob/master/playbooks/roles/edx_django_service/defaults/main.yml#L182-L222

as they get merged into the edx_django_service_config here:
https://github.com/edx/configuration/blob/master/playbooks/roles/edx_django_service/defaults/main.yml#L226

analyticsdataserver/settings/base.py
```diff
@@ -42,16 +42,31 @@
 
 ########## DATABASE CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#databases
+DEFAULT_MYSQL_OPTIONS = {
```
coryleeio: Can you copy CACHES into here from local.py so that it will be defined?

analyticsdataserver/settings/base.py
```diff
@@ -42,16 +42,31 @@
```
coryleeio: We are ignoring these ones:

  MEDIA_STORAGE_BACKEND # not used
  STATICFILES_STORAGE   # not used
  SOCIAL_AUTH_EDX_OIDC_KEY # not used
  SOCIAL_AUTH_EDX_OIDC_SECRET # not used
  SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY # not used
  SOCIAL_AUTH_EDX_OIDC_URL_ROOT # not used
  SOCIAL_AUTH_EDX_OIDC_PUBLIC_URL_ROOT # not used
  SOCIAL_AUTH_REDIRECT_IS_HTTPS # not used
  SOCIAL_AUTH_EDX_OIDC_LOGOUT_URL # not used
  SOCIAL_AUTH_EDX_OIDC_ISSUER # not used
  SOCIAL_AUTH_EDX_OAUTH2_KEY # not used
  SOCIAL_AUTH_EDX_OAUTH2_SECRET # not used
  SOCIAL_AUTH_EDX_OAUTH2_ISSUER # not used
  SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT # not used
  SOCIAL_AUTH_EDX_OAUTH2_LOGOUT_URL # not used
  BACKEND_SERVICE_EDX_OAUTH2_KEY # not used
  BACKEND_SERVICE_EDX_OAUTH2_SECRET # not used
  BACKEND_SERVICE_EDX_OAUTH2_PROVIDER_URL # not used
  EXTRA_APPS # not used
  EDX_DRF_EXTENSIONS # not used
  SESSION_EXPIRE_AT_BROWSER_CLOSE # not used
  API_ROOT # not used

because they are not used, is that correct?

syedimranhassan: I have added these variables. These variables are overriding in meta/main.yml.

analyticsdataserver/settings/base.py
syedimranhassan: I have moved databases to local.py and copy caches from local.py to base.py

analyticsdataserver/settings/local.py
```diff
@@ -22,22 +22,28 @@
 
 ########## DATABASE CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#databases
+DEFAULT_MYSQL_OPTIONS = {
+    'connect_timeout': 10,
+    'init_command': "SET sql_mode='STRICT_TRANS_TABLES'"
+}
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
```
coryleeio: Same thing here, lets not change their local.py

analyticsdataserver/settings/base.py
```diff
@@ -39,21 +39,6 @@
 MANAGERS = ADMINS
 ########## END MANAGER CONFIGURATION
 
-
-########## DATABASE CONFIGURATION
-# See: https://docs.djangoproject.com/en/dev/ref/settings/#databases
-DATABASES = {
-    'default': {
-        'ENGINE': 'django.db.backends.',
-        'NAME': '',
```
coryleeio: Lets put CACHES and DATABASES in the base.py, and not change the local.py for now.

PR #274, opened 2019-07-31
EDUCATOR-4533 initial implementation of simple get

Andytr1:
We would like to be able to pull data from module_engagement. The main purpose of this work is to have an initial endpoint that provides simple data.

### DIFF
--- a/analytics_data_api/v0/models.py
+++ b/analytics_data_api/v0/models.py
@@ -493,6 +493,10 @@ class ModuleEngagementTimelineManager(models.Manager):
 
         return full_timeline
 
+    def get_simple_data_for_all_students(self, course_id):
+        return ModuleEngagement.objects.all().filter(course_id=course_id) \
+            .order_by('username')
+
 
 class ModuleEngagement(BaseCourseModel):
     """User interactions with entities within the courseware."""
--- a/analytics_data_api/v0/urls/courses.py
+++ b/analytics_data_api/v0/urls/courses.py
@@ -16,6 +16,7 @@ COURSE_URLS = [
     ('problems_and_tags', views.ProblemsAndTagsListView, 'problems_and_tags'),
     ('videos', views.VideosListView, 'videos'),
     ('reports/(?P<report_name>[a-zA-Z0-9_]+)', views.ReportDownloadView, 'reports'),
+    ('user_engagement', views.UserEngagementView, 'user_engagement'),
 ]
 
 urlpatterns = []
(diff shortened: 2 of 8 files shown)


Andytr1:
jenkins run travis

analytics_data_api/v0/models.py
```diff
@@ -493,6 +493,12 @@ def get_timeline(self, course_id, username):
 
         return full_timeline
 
+    def get_simple_data_for_all_students(self, course_id):
+        queryset = ModuleEngagement.objects.all().filter(course_id=course_id) \
+            .order_by('username')
+
+        return [entry for entry in queryset]
```
mulby: I think you can/should just return the queryset here instead of resolving it directly to a python list.

related: I would recommend putting pagination of some kind on this API. IIRC that's easier to do if this is a queryset instead of a list.

analytics_data_api/v0/urls/courses.py
```diff
@@ -16,6 +16,7 @@
     ('problems_and_tags', views.ProblemsAndTagsListView, 'problems_and_tags'),
     ('videos', views.VideosListView, 'videos'),
     ('reports/(?P<report_name>[a-zA-Z0-9_]+)', views.ReportDownloadView, 'reports'),
+    ('intervention', views.CourseInterventionView, 'intervention'),
```
mulby: All of this data can be used by an instructor for the purpose of intervention. I would recommend naming this something that tells the API consumer more about what the data itself is.

Andytr1: would 'intervention_engagements'  be better ?

mulby: How about just "user_engagement"?

analytics_data_api/v0/models.py
Andytr1: I don't think pagination makes sense in our use case as we our exporting all data into .csv, not displaying it on screen.

davestgermain: There's no harm in returning the queryset here. In fact it would be clearer.

analytics_data_api/v0/views/courses.py
```diff
@@ -810,3 +814,52 @@ def get(self, _request, course_id, report_name):
             return Response(response)
         else:
             raise ReportFileNotFoundError(course_id=course_id, report_name=report_name)
+
+
+# pylint: disable=abstract-method
```
matthugs: I was confused by this, so I pulled to check whether it was necessary. It looks like `make quality` runs cleanly without this line; could you remove it?

PR #278, opened 2019-08-15
EDUCATOR-4534 Endpoint to get aggregate data

Andytr1:
EDUCATOR-4534 - get aggregated data

### DIFF
--- a/analytics_data_api/v0/models.py
+++ b/analytics_data_api/v0/models.py
@@ -3,7 +3,8 @@ from itertools import groupby
 
 from django.conf import settings
 from django.db import models
-from django.db.models import Count, Sum
+from django.db.models import Count, Sum, When, IntegerField, Case, Max
+from django.utils.timezone import now
 # some fields (e.g. Float, Integer) are dynamic and your IDE may highlight them as unavailable
 from elasticsearch_dsl import Date, DocType, Float, Integer, Q, String  # pylint: disable=no-name-in-module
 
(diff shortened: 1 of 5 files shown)


analytics_data_api/v0/models.py
```diff
@@ -497,6 +497,44 @@ def get_simple_data_for_all_students(self, course_id):
...
+                problems_attempts_last_week=Sum(Case(When(entity_type='problem', event='attempted',
+                                                          created__gt=seven_days_ago, then='count'),
+                                                     output_field=IntegerField()))) \
+            .annotate(
+                forum_posts_overall=Sum(Case(When(entity_type='discussion', then='count'),
+                                             output_field=IntegerField()))) \
+            .annotate(
+                forum_posts_last_week=Sum(Case(When(entity_type='discussion', created__gt=seven_days_ago, then='count'),
+                                               output_field=IntegerField()))) \
+            .annotate(
```
matthugs: I think annotate can take a single dict, which would let us reduce this down to one call & reduce some grossness.

Also I feel like we probably don't have linting running in CI if this is building green. Could you check this with `make quality` after `make requirements` (you may want to do this in a virtualenv)

Andytr1: linting is running. I ran make quality and everything was green.

Andytr1: Here is the SQL query generated - it appears to be a standard aggregate. Not sure what else can be done to optimize it:
SELECT module_engagement.username, SUM(CASE WHEN module_engagement.entity_type = 'video' THEN count ELSE NULL END) AS videos_overall, 
SUM(CASE WHEN (module_engagement.created > STR_TO_DATE('2019-08-07', '%Y-%m-%d') AND module_engagement.entity_type = 'video') THEN count ELSE NULL END) 
AS videos_last_week, 
SUM(CASE WHEN module_engagement.entity_type = 'problem' THEN count ELSE NULL END) AS problems_overall, 
SUM(CASE WHEN (module_engagement.created > STR_TO_DATE('2019-08-07', '%Y-%m-%d') AND module_engagement.entity_type = 'problem') THEN count ELSE NULL END) 
AS problems_last_week, 
SUM(CASE WHEN (module_engagement.event = 'completed' AND module_engagement.entity_type = 'problem') THEN count ELSE NULL END) 
AS correct_problems_overall, 
SUM(CASE WHEN (module_engagement.created > STR_TO_DATE('2019-08-07', '%Y-%m-%d') AND module_engagement.event = 'completed' 
AND module_engagement.entity_type = 'problem') THEN module_engagement.count ELSE NULL END) 
AS correct_problems_last_week, 
SUM(CASE WHEN (module_engagement.event = 'attempted' AND module_engagement.entity_type = 'problem') THEN count ELSE NULL END) AS problems_attempts_overall, 
SUM(CASE WHEN (module_engagement.created > STR_TO_DATE('2019-08-07', '%Y-%m-%d') AND module_engagement.event = 'attempted'
AND module_engagement.entity_type = 'problem') THEN count ELSE NULL END) AS problems_attempts_last_week, 
SUM(CASE WHEN module_engagement.entity_type = 'discussion' THEN count ELSE NULL END) AS forum_posts_overall, 
SUM(CASE WHEN (module_engagement.created > STR_TO_DATE('2019-08-07', '%Y-%m-%d') AND module_engagement.entity_type = 'discussion') THEN count ELSE NULL END) 
AS forum_posts_last_week, MAX(module_engagement.created) AS date_last_active 
FROM module_engagement
WHERE module_engagement.course_id = 'course-v1:seakademieX+1+2T2018'
GROUP BY module_engagement.username;

matthugs: The EXPLAIN info we got from this (out of band) was actually surprisingly uninstructive. Marking as resolved.

matthugs: Or, wait, the core of this comment was that we should switch to object-arg'd `annotate`. Let's still do that

Andytr1: I am having difficulty tracking down object-arg'd annotate. Can you link to a doc or existing code base.

matthugs: Based purely on the fact that there are stars in front of the kwargs in https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.query.QuerySet.annotate, I suspect you can provide any number of kwargs to a single call to annotate, which is really what I'm after here.

davestgermain: I concur: a single `annotate()` call should be nicer.

Andytr1:
> I'm concerned that this is going to be a painfully slow query in production. Have you tried running it?

It takes about 1.5-2 minutes on a read replica. However, it only limits me to 1000 records. I think that's reasonable, given that the user is downloading a file.

davestgermain:
> It takes about 1.5-2 minutes on a read replica. However, it only limits me to 1000 records. I think that's reasonable, given that the user is downloading a file.

I hope you mean *seconds*...

PR #296, opened 2020-02-14
Updated the dependencies - BOM-1276

mraarif:
Relevant JIRA issue can be found [here](https://openedx.atlassian.net/browse/BOM-1276).


requirements/base.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    make upgrade
+#    pip-compile --output-file=requirements/base.txt requirements/base.in
```
jmbowman: We should set `CUSTOM_COMPILE_COMMAND` in `make upgrade` so this stays as it was.  Example: https://github.com/edx/edx-platform/blob/master/Makefile#L84

PR #335, opened 2020-03-12
[BD] Fix RemovedInDjango20Warning: Specifying a namespace in django.conf.urls.include() without providing an app_name is deprecated.

morenol:
## Description
[BD] Fix RemovedInDjango20Warning: Specifying a namespace in django.conf.urls.include() without providing an app_name is deprecated.

https://docs.djangoproject.com/en/dev/releases/1.9/#passing-a-3-tuple-or-an-app-name-to-include

Related to https://openedx.atlassian.net/browse/BOM-1359

## Reviewers
- [ ] @andrey-canon
- [x] ready for edX review.
- [ ] @jmbowman

### DIFF
--- a/analytics_data_api/v0/urls/course_summaries.py
+++ b/analytics_data_api/v0/urls/course_summaries.py
@@ -4,6 +4,8 @@ from django.conf.urls import url
 
 from analytics_data_api.v0.views import course_summaries as views
 
+app_name = 'course_summaries'
+
 urlpatterns = [
     url(r'^course_summaries/$', views.CourseSummariesView.as_view(), name='course_summaries'),
 ]
--- a/analytics_data_api/urls.py
+++ b/analytics_data_api/urls.py
@@ -3,8 +3,10 @@ from __future__ import absolute_import
 from django.conf.urls import include, url
 from rest_framework.urlpatterns import format_suffix_patterns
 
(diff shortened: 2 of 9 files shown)


edx-webhook:
Thanks for the pull request, @morenol! I've created [OSPR-4242](https://openedx.atlassian.net/browse/OSPR-4242) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

analytics_data_api/v0/urls/course_summaries.py
```diff
@@ -4,6 +4,8 @@
 
 from analytics_data_api.v0.views import course_summaries as views
 
+app_name = 'analytics_data_api.v0.course_summaries'
```
jmbowman: Note that this is only used as an optional prefix when looking up URLs by name.  We haven't used the ones previously declared in the `include()` calls anywhere, so to some extent it doesn't matter what the value is.  But these are long enough that I'm not sure we'd ever want to use them; `reverse("analytics_data_api.v0.course_summaries:course_summaries")` is a little verbose.

Maybe we'd just be fine using values like `app_name = 'course_summaries'`, since that's effectively what we had before?

PR #340, opened 2020-03-17
Changes for Devstack

morenol:
Add devstack settings for the docker devstack.

This is related to: https://github.com/edx/configuration/pull/5704/files and to https://github.com/edx/devstack/pull/480

Reviewers
- [ ] @andrey-canon
- [x] Ready for edX review.
- [ ] @jmbowman

### DIFF
--- a/analyticsdataserver/settings/devstack.py
+++ b/analyticsdataserver/settings/devstack.py
@@ -1,0 +1,27 @@
+"""Devstack settings."""
+
+from analyticsdataserver.settings.local import *
+
+ALLOWED_HOSTS += ['edx.devstack.analyticsapi']
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.mysql',
+        'NAME': 'analytics-api',
+        'USER': 'api001',
+        'PASSWORD': 'password',
+        'HOST': 'edx.devstack.mysql',
+        'PORT': '3306',
+    },
+    'analytics': {
+        'ENGINE': 'django.db.backends.mysql',
+        'NAME': 'reports',
+        'USER': 'reports001',
+        'PASSWORD': 'password',
+        'HOST': 'edx.devstack.mysql',
+        'PORT': '3306',
+    }
+}
+
+ELASTICSEARCH_LEARNERS_HOST = "edx.devstack.elasticsearch"
+LMS_BASE_URL = "http://edx.devstack.lms:18000/"


edx-webhook:
Thanks for the pull request, @morenol! I've created [OSPR-4268](https://openedx.atlassian.net/browse/OSPR-4268) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

morenol:
@jmbowman I think that we can start to merge this kind of PRs in order to get a devstack with analytics-dashboard since this is not blocked by the python2.7 PRs

analyticsdataserver/settings/devstack.py
```diff
@@ -0,0 +1,24 @@
+from .local_mysql import *
+
+ALLOWED_HOSTS += ['edx.devstack.analyticsapi']
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.mysql',
```
jmbowman: Why do these settings use MySQL for default and SQLite for analytics, but `local_mysql.py` does the opposite?  Is there a good reason not to use MySQL for both databases?

morenol: I made this because `./manage.py migrate --noinput --run-syncdb --database=analytics` fails when the analytics database is mysql.

This is the error that is thrown by django: ```django.db.utils.IntegrityError: (1215, 'Cannot add foreign key constraint')```

This is the problematic migration: https://github.com/edx/edx-enterprise-data/blob/master/enterprise_data/migrations/0012_auto_20180831_1930.py


This is the error that reports mysql in the devstack: 
```
Error in foreign key constraint of table reports/#sql-1_27:
FOREIGN KEY (`enterprise_user_id_id`) REFERENCES `enterprise_user` (`enterprise_user_id`):
Cannot find an index in the referenced table where the
referenced columns appear as the first columns, or column types
in the table and the referenced table do not match for constraint.
Note that the internal storage type of ENUM and SET changed in
tables created with >= InnoDB-4.1.12, and such columns in old tables
cannot be referenced by such columns in new tables.
See http://dev.mysql.com/doc/refman/5.6/en/innodb-foreign-key-constraints.html
for correct foreign key definition.```

I think that we have not experienced this error before because in the tests it is used the local.py (which uses sqlite for both databases)

morenol: And I think that in prod we don't have that problem because we never run `manage.py migrate --noinput --run-syncdb --database=analytics` and instead, we populate that database from the edx-analytics-pipeline service.

jmbowman: I vaguely remember discussing that problem before in an edx-enterprise-data PR; can we just fix the migration?  Seems to be a flaw from https://github.com/edx/edx-enterprise-data/commit/61f6896f546627b5e1e021b134356d2ed6dc3966 .

morenol: I was able to run the migration doing this:

Run the migrations until the migration before the problematic migration, i.e:

`python manage.py migrate enterprise_data 0011 --database analytics`

Run this on the mysql container:

`ALTER IGNORE TABLE enterprise_user ADD UNIQUE (enterprise_user_id);`

And then run the rest of the migrations, i.e:

`python manage.py migrate enterprise_data --database analytics`

morenol: It seems that mysql is complaining because we are trying to create a foreign key to a non unique field https://stackoverflow.com/questions/18435065/foreign-key-to-non-primary-key/18435114#18435114

morenol: It seems that in the model definition that should be unique, I will create a PR to fix that.

morenol:
This is now configured with mysql for both databases, this is the PR in edx-enterprise-data:

https://github.com/edx/edx-enterprise-data/pull/191

PR #353, opened 2020-03-25
Fix type checks on elastisearch responses to fix learner dashboard.

macdiesel:


analytics_data_api/v0/connections.py
```diff
@@ -40,7 +40,7 @@ def perform_request(self, method, url, params=None, body=None, timeout=None, ign
         and the default is 70 seconds.
         See: https://github.com/boto/boto/blob/develop/boto/connection.py#L533
         """
-        if not isinstance(body, six.string_types):
+        if not isinstance(body, str) or not isinstance(body, bytes):
```
jmbowman: Don't you want `and` instead?  As written I think this is always true.

macdiesel: :doh: Fixed.

PR #355, opened 2020-04-01
Constraining edx-enterprise-data to 1.3.15

ziafazal:
Constraining edx-enterprise-data  to 1.3.15 until we have next release with fix for ENT-2717 https://github.com/edx/edx-enterprise-data/pull/192


morenol:
I tested this locally, and it looks like this not solves the problem, I think that the problem is related to the python3 switch instead of any other package change

macdiesel:
I'm going to close this PR. If this issue is still ongoing please open a new PR and a DESupport ticket.

PR #361, opened 2020-04-03
[BD] Upgrade django to Django 2.2

morenol:
## Description

Upgrade to django 2.2

## Reviewers
- [ ] @andrey-canon
- [x] Ready for edX review
- [ ] @jmbowman


morenol:
Current warnings in Django 2.2 version: 383
From which 318 are ResourceWarning: unclosed <socket... with the elasticsearch test container.\

morenol:
@jmbowman Here I removed the constraint to django-waffle which is meant to remove the support for django <2.2. Should I remove the other tests? or should I add the constraint to django-waffle?

jmbowman:
@morenol - no, let's keep those for now.  Can delete after this has been running on prod for a little while without needing to roll it back.

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5387](https://openedx.atlassian.net/browse/OSPR-5387) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #367, opened 2020-04-16
Updated Python and Django in Travis and Tox

UsamaSadiq:
**Issue:** [BOM-1530](https://openedx.atlassian.net/browse/BOM-1530)

### Description
- Added `python3.8` and `django3.0` in tox.
- Removed `django111`, `django2.0` and `django2.1` from Travis.
- Added `python3.8` in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
-python:
-  - 3.5
+
 branches:
   only:
     - master
@@ -23,21 +22,26 @@ script:
 
 matrix:
   include:
-    - env:
-        TESTNAME=quality
-        TARGETS="quality"
-    - env:
-        TESTNAME=test-python
-        TARGETS="test"
+    - python: 3.5
+      env:
+        TESTNAME=quality-python-3.5
+        TARGETS="PYTHON_ENV=py35 quality"
+    - python: 3.5
+      env:
+        TESTNAME=test-python-3.5
+        TARGETS="PYTHON_ENV=py35 test"
       after_success:
         - docker exec analytics_api_testing /edx/app/analytics_api/analytics_api/.travis/run_coverage.sh
         - codecov --disable pycov
-    - env:
-        TESTNAME=test-python-django20
-        TARGETS="DJANGO_VERSION=django20 test"
-    - env:
-        TESTNAME=test-python-django21
-        TARGETS="DJANGO_VERSION=django21 test"
-    - env:
-        TESTNAME=test-python-django111
-        TARGETS="DJANGO_VERSION=django111 test"
+
+    - python: 3.8
+      env:
+        TESTNAME=quality-python-3.8
+        TARGETS="PYTHON_ENV=py38 quality"
+    - python: 3.8
+      env:
+        TESTNAME=test-python-3.8
+        TARGETS="PYTHON_ENV=py38 test"
+
+  allow_failures:
+    - python: 3.8
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -13,23 +13,11 @@
 # Use latest Django LTS version
 Django>=2.2,<2.3.0
 
-# django-model-utils version 4.0.0 dropped support to python 2.7	
-django-model-utils<4.0.0	
-
 # django-storages version 1.9 drops support for boto storage backend.
 django-storages<1.9                 # BSD
-
-# django-waffle version 0.19.0 dropped support to Django 2.0 and 2.1, see https://github.com/django-waffle/django-waffle/blob/master/CHANGES
-django-waffle<0.19.0
 
 # django22 tests are failing. Constraint taken from https://github.com/edx/edx-drf-extensions version 5.0.2
 drf-jwt<1.15.0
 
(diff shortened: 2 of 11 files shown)


.travis.yml
```diff
@@ -33,11 +34,5 @@ matrix:
...
-        TESTNAME=test-python-django20
-        TARGETS="DJANGO_VERSION=django20 test"
-    - env:
-        TESTNAME=test-python-django21
-        TARGETS="DJANGO_VERSION=django21 test"
-    - env:
-        TESTNAME=test-python-django111
-        TARGETS="DJANGO_VERSION=django111 test"
+        TESTNAME=test-python-django22
+        TARGETS="DJANGO_VERSION=django22 test"
```
morenol, line 38: Hi, I think that you should put in TARGETS="PYTHON_ENV=py35" in order to use that version because all of these tests run on the container and not in the travis environment.

UsamaSadiq: Hi @morenol, thanks for the advice. I've made the changes according to your suggestion. Could you have a look again and see if I am missing anything else? Thanks.

.travis.yml
```diff
@@ -23,21 +22,35 @@ script:
...
         - docker exec analytics_api_testing /edx/app/analytics_api/analytics_api/.travis/run_coverage.sh
         - codecov --disable pycov
     - env:
-        TESTNAME=test-python-django20
-        TARGETS="DJANGO_VERSION=django20 test"
-    - env:
-        TESTNAME=test-python-django21
-        TARGETS="DJANGO_VERSION=django21 test"
+        TESTNAME=test-python-3.5-django22
+        TARGETS="PYTHON_ENV=py35 DJANGO_VERSION=django22 test"
```
morenol, line 38: I think that this is repeating:

```yaml
    - python: 3.5
      env:
        TESTNAME=test-python-3.5
        TARGETS="PYTHON_ENV=py35 test"
since django22 is the default version in the tests (See makefile default DJANGO_VERSION).

File: .travis.yml [hunk 3]
```diff
@@ -23,21 +22,35 @@ script:
...
+      env:
+        TESTNAME=test-python-3.8
+        TARGETS="PYTHON_ENV=py38 test"
+      after_success:
+        - docker exec analytics_api_testing /edx/app/analytics_api/analytics_api/.travis/run_coverage.sh
+        - codecov --disable pycov
     - env:
-        TESTNAME=test-python-django111
-        TARGETS="DJANGO_VERSION=django111 test"
+        TESTNAME=test-python-3.8-django22
```
morenol, line 52: Similarly, I think that this is repeating the previous test

.travis.yml
```diff
@@ -23,21 +22,35 @@ script:
...
+
+    - python: 3.8
+      env:
+        TESTNAME=quality-python-3.8
+        TARGETS="PYTHON_ENV=py38 quality"
+    - python: 3.8
+      env:
+        TESTNAME=test-python-3.8
+        TARGETS="PYTHON_ENV=py38 test"
+      after_success:
```
morenol, line 48: I am not sure, but I think that we should not run this at the same time for both versions of python.

.travis.yml
UsamaSadiq: I was thinking the same but wanted to confirm, thanks.

.travis.yml
UsamaSadiq: We need to add python3.8 support for tests too so how should we go about it then?

morenol: The test will be running anyway, I mean that I think that we should remove the after_success section in order to upload the coverage only once to the codecov.

requirements/constraints.txt
```diff
@@ -11,25 +11,13 @@
 # TODO: Many pinned dependencies should be unpinned and/or moved to this constraints file.
 
 # Use latest Django LTS version
-Django>=2.2,<2.3.0
-
-# django-model-utils version 4.0.0 dropped support to python 2.7	
-django-model-utils<4.0.0	
+Django>=2.2,<3.1
```
morenol, line 14: I think that if we run make upgrade, then the new django version will be 3.0.X is this something expected?

Also, if we change the constraints I think that we should run make upgrade

UsamaSadiq: Yes, I was working on Travis changes first. I'll run make upgrade first before getting the reviews on it. 
Thanks for your help.

requirements/constraints.txt
```diff
@@ -11,25 +11,13 @@
 # TODO: Many pinned dependencies should be unpinned and/or moved to this constraints file.
 
 # Use latest Django LTS version
-Django>=2.2,<2.3.0
-
-# django-model-utils version 4.0.0 dropped support to python 2.7	
-django-model-utils<4.0.0	
+Django>=2.2,<2.3.1
```
jmbowman, line 14: Not that it should matter, but why the change in upper limit?

morenol:
> does this now fit with the changes you've been making here?
@jmbowman This seems good to me, my only concern is in the Django upper limit change

requirements/constraints.txt
UsamaSadiq: Changed back to previous `Django>=2.2,<2.3.0`.

PR #372, opened 2020-05-07
add travis dockerhub config

fredsmith:


.travis.yml
```diff
@@ -1,47 +1,38 @@
...
-        TARGETS="PYTHON_ENV=py38 test"
-
+  - python: 3.5
+    env: TESTNAME=quality-python-3.5 TARGETS="PYTHON_ENV=py35 quality"
+  - python: 3.5
+    env: TESTNAME=test-python-3.5 TARGETS="PYTHON_ENV=py35 test"
+    after_success:
+    - docker exec analytics_api_testing /edx/app/analytics_api/analytics_api/.travis/run_coverage.sh
+    - codecov --disable pycov
+  - python: 3.8
```
adzuci, line 24: I'm curious why we're running 3.8 or if we should be running other versions, but I think it should stay as is for now.

PR #374, opened 2020-05-11
move dockerfile to python 3

fredsmith:


coryleeio:
I think this is fine, but I would add this:

RUN ln -s /usr/bin/pip3 /usr/bin/pip
RUN ln -s /usr/bin/python3 /usr/bin/python

so you dont have to specify python3 and pip3

PR #394, opened 2020-07-29
Add edx-enterprise-data installation instructions

raq929:
Instructions for installing edx-enterprise-data and links to the frontend enterprise app.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,6 +5,8 @@ This repository includes the Django server for the API as well as the
 API package itself. The client is hosted at
 https://github.com/edx/edx-analytics-data-api-client.
 
+The enterprise client is hosted at https://github.com/edx/frontend-app-admin-portal.
+
 License
 -------
 
@@ -15,14 +17,15 @@ Please see ``LICENSE.txt`` for details.
 
 Getting Started
 ---------------
+#. Create a virtual environment and activate it.
 
-1. Install the requirements:
+#. Install the requirements:
 
    ::
 
        $ make develop
 
-2. Setup the databases:
+#. Setup the databases:
 
    ::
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -45,21 +48,32 @@ Getting Started
 
       $ make test.run_elasticsearch
 
-3. Create a user and authentication token. Note that the user will be
+#. Create a user and authentication token. Note that the user will be
    created if one does not exist.
 
    ::
 
        $ ./manage.py set_api_key <username> <token>
```
binodpant, line 56: I always wondered what this token here should be, maybe can add a line? I have been entering garbage values for it like '1alkdjflsd'

README.rst
```diff
@@ -45,21 +48,32 @@ Getting Started
...
 
+Development with edx-enterprise-data
+------------------------------------
+If you need to make changes to ``edx-enterprise-data`` and have them reflected when you run the ``edx-analytics-data-api`` server,
+you can follow these steps. If you do not intend to make changes to ``edx-enterprise-data``, you can skip this section.
+
+#. Recommended: Install this repo into a subfolder of your working directory. Within that subfolder create an ``src`` folder.
+#. Clone the `edx-enterprise-data <https://github.com/edx/edx-enterprise-data>`_ repo into the ``src`` folder.
+#. ``cd`` into your ``edx-data-analytics-api`` folder and activate your virtualenv.
+#. Run ``/src/edx-enterprise-data``.
```
binodpant, line 74: is this `/edx/src/edx-enterprise-data` when in the shell?

README.rst
raq929: Ha, I'm not exactly sure what it is myself; I assume it's a token for API access? imho we should be generating that ourselves and giving it to the user, not the other way around...

README.rst
raq929: Not for me. My file structure is:
```
edx
  |
  analytics
    |
    edx-analytics-data-api
    src
        |
        edx-enterprise-data
```

Oh, but that should be a pip install command. Let me update.

PR #395, opened 2020-07-30
Upgrade requirements

raq929:
Upgrades requirements.

I did not include the major version upgrade to cryptography, as it contains breaking changes.

Also creates a `make help` command as the default make command and adds docstrings to the makefile.

### DIFF
--- a/requirements/pip_tools.txt
+++ b/requirements/pip_tools.txt
@@ -5,5 +5,8 @@
 #    make upgrade
 #
 click==7.1.2              # via pip-tools
-pip-tools==4.5.1          # via -r requirements/pip_tools.in
+pip-tools==5.3.0          # via -r requirements/pip_tools.in
 six==1.15.0               # via -r requirements/pip_tools.in, pip-tools
+
+# The following packages are considered to be unsafe in a requirements file:
+# pip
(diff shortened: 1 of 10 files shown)


requirements/pip_tools.txt
```diff
@@ -5,5 +5,8 @@
 #    make upgrade
 #
 click==7.1.2              # via pip-tools
-pip-tools==4.5.1          # via -r requirements/pip_tools.in
```
binodpant, line 8: there is a major upgrade here.. will be good to ensure everything still works, not sure what exaclt that means but maybe running some make targets, did we try make test etc?

requirements/test.txt
```diff
@@ -76,7 +77,7 @@ pymongo==3.10.1           # via edx-opaque-keys
 pyparsing==2.4.7          # via packaging
 pytest-cov==2.10.0        # via -r requirements/test.in
 pytest-django==3.9.0      # via -r requirements/test.in
-pytest==5.4.3             # via pytest-cov, pytest-django
```
binodpant, line 79: this gave me lots of trouble in edx-enterprise, will be good to confirm if builds pass with this, if so all good! I think edx-enterprise upgrade of this is still open

requirements/pip_tools.txt
raq929: I'll try make test. I upgraded it because I couldn't run `make upgrade` without upgrading it. 😝

PR #400, opened 2020-09-01
updated the Dockerfile to use python 3.8

mraarif:
Relevant JIRA issue here: https://openedx.atlassian.net/browse/BOM-2005

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,16 @@
 FROM ubuntu:xenial as app
 
 RUN apt update && \
-  apt install -y git-core language-pack-en python3.5 python3-pip python3-dev libmysqlclient-dev libffi-dev libssl-dev build-essential gettext openjdk-8-jdk && \
-  pip3 install --upgrade pip setuptools && \
+  apt-get install -y software-properties-common && \
+  apt-add-repository -y ppa:deadsnakes/ppa && apt-get update && \
+  apt install -y git-core language-pack-en python3.8-dev python3.8-venv libmysqlclient-dev libffi-dev libssl-dev build-essential gettext openjdk-8-jdk && \
   rm -rf /var/lib/apt/lists/*
 
+ENV VIRTUAL_ENV=/venv
+RUN python3.8 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
-RUN ln -s /usr/bin/pip3 /usr/bin/pip
-RUN ln -s /usr/bin/python3 /usr/bin/python
-
+RUN pip install pip==20.2.3 setuptools==50.3.0
 
 RUN locale-gen en_US.UTF-8
 ENV LANG en_US.UTF-8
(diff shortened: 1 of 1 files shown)


mraarif:
changes tested by building an image locally.

Dockerfile
```diff
@@ -1,11 +1,13 @@
 FROM ubuntu:xenial as app
 
 RUN apt update && \
-  apt install -y git-core language-pack-en python3.5 python3-pip python3-dev libmysqlclient-dev libffi-dev libssl-dev build-essential gettext openjdk-8-jdk && \
-  pip3 install --upgrade pip setuptools && \
+  apt-get install -y software-properties-common && \
+  apt-add-repository -y ppa:deadsnakes/ppa && apt-get update && \
+  apt install -y git-core language-pack-en python3.5 python3-pip python3-dev python3.8-dev python3.8-distutils libmysqlclient-dev libffi-dev libssl-dev build-essential gettext openjdk-8-jdk && \
```
jmbowman, line 6: It looks like this file was added for future use in prod and decentralized devstack; can we skip installation of `python3.5` and `python3-dev`?  I don't think anything would still be using them after this update.

Dockerfile
```diff
@@ -18,7 +19,7 @@ ENV ANALYTICS_API_CFG /edx/etc/analytics_api.yml
 
 WORKDIR /edx/app/analytics_api
 COPY requirements /edx/app/analytics_api/requirements
-RUN pip3 install -r requirements/production.txt
+RUN python3.8 -m pip install -r requirements/production.txt
```
jmbowman, line 22: With the environment variables set above to activate the virtualenv, this can just be the simpler `RUN pip install ...` (although what's here now should work also).

Dockerfile
```diff
@@ -28,5 +29,5 @@ USER app
 COPY . /edx/app/analytics_api
 
 FROM app as newrelic
-RUN pip3 install newrelic
+RUN python3.8 -m pip install newrelic
```
jmbowman, line 32: Same here.

Dockerfile
```diff
@@ -1,14 +1,15 @@
 FROM ubuntu:xenial as app
 
 RUN apt update && \
-  apt install -y git-core language-pack-en python3.5 python3-pip python3-dev libmysqlclient-dev libffi-dev libssl-dev build-essential gettext openjdk-8-jdk && \
-  pip3 install --upgrade pip setuptools && \
+  apt-get install -y software-properties-common && \
+  apt-add-repository -y ppa:deadsnakes/ppa && apt-get update && \
+  apt install -y git-core language-pack-en python3-pip python3.8-dev python3.8-venv libmysqlclient-dev libffi-dev libssl-dev build-essential gettext openjdk-8-jdk && \
```
jmbowman, line 6: Now that we're using the virtualenv, I don't think you should need `python3-pip`.

jmbowman (approved):
Looks good other than the noted minor optimizations.  Just make sure you can still successfully build the image with this.

Dockerfile
mraarif: done

mraarif:
I was able to build an image with these changes.

macdiesel (approved):

PR #425, opened 2021-01-29
DESUPPORT-721: Add NewRelic tracing to the course summaries GET call

bmedx:


macdiesel (approved):

doctoryes (approved):
Doesn't appear that the edx-platform way of implementing these traces would add any value:
https://github.com/edx/edx-django-utils/blob/master/edx_django_utils/monitoring/internal/transactions.py#L44-L63
so 👍🏼

PR #454, opened 2021-07-14
fix: add required header when making ES request

Zacharis278:
@edx/masters-devs-cosmonauts 

No good way to test this locally so this could be whack-a-mole fix.

This is the function being called: https://github.com/boto/boto/blob/develop/boto/connection.py#L1065


ashultz0 (approved):
certainly plausible

alangsto (approved):

PR #455, opened 2021-07-14
fix: match existing index types

Zacharis278:
@edx/masters-devs-cosmonauts 

Fixes a serialization bug where our mappings didn't match the pipeline output exactly.

The pipeline code doesn't have a type definition to write a short so these are just getting output as an integer. Seemed easier to update this on our end to just read int instead.


schenedx (approved):
👍

PR #458, opened 2021-07-16
fix: return List instead of ES response object

Zacharis278:
@edx/masters-devs-cosmonauts 

Somewhat of a shot in the dark but the splunk logs seem to indicate a different object being returned here than I'm seeing locally 🤯 .

### DIFF
--- a/analytics_data_api/v0/documents.py
+++ b/analytics_data_api/v0/documents.py
@@ -1,8 +1,11 @@
+import logging
 
 from django.conf import settings
 from elasticsearch_dsl import Date, Document, Float, Integer, Keyword, Q
 
 from analytics_data_api.constants import learner
+
+logger = logging.getLogger(__name__)
 
 
 class RosterUpdate(Document):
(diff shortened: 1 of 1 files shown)


schenedx (approved):
Please revert once you found the issues.

analytics_data_api/v0/documents.py
```diff
@@ -146,7 +149,13 @@ def get_users_in_course(
             }
             for sort_policy in sort_policies
         ])
-        return search_request.execute()
+        res = search_request.execute()
+        # debugging, to be removed
+        logger.warning(res.__class__.__name__)
+        items = res.hits
+        logger.warning(items[0])
```
ashultz0, line 156: if it's a different object than you expect are you sure res hits exists and is an array and has something in it?

does the caller do hits and hits[0] and isn't crashing so this must be ok?

ashultz0 (approved):

analytics_data_api/v0/documents.py
Zacharis278: I think the questionable objects are within `hits`.  I'd expect this class RosterEntry but I'm seeing a generic Hit object instead (I think).

PR #460, opened 2021-07-23
feat: support for aliased index

Zacharis278:
@edx/masters-devs-cosmonauts 

The root of our issues was that the INDEX setting passed in from the environment was actually an alias to an index and not an actual index in the AWS environment. The local setup however did not reflect that, which led to different behavior when this project is run locally.

1. This resolves issues with the updated elasticsearch-dsl not supporting queries using an alias as a search target. Well, the queries work fine, mapping back to a persistent python type does not.
2. This PR also updates the local setup to reflect a cluster intended to be accessed to using an alias.
3. I've updated the name of the INDEX django setting to be more clear that it is an alias to an index.

why do we do this?
https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/insights/module_engagement.py#L1361

edx-internal PR: https://github.com/edx/edx-internal/pull/5296

Final note. I'm going to open a ticket to better document how and why ES is set up this way


schenedx (approved):
When are you planning to remove the log statements which helped you debug?

ashultz0 (approved):
weird find

it feels like there is a missing make ES less horrifying layer for the future if we're going to write a lot of analytics into it

or a replacement for ES which is meant as an analytics store not as the monster ES is now

Zacharis278:
> When are you planning to remove the log statements which helped you debug?

I think I got them all

PR #469, opened 2021-08-06
temp: just checking newly added test with pyjwt2.0

awais786:
Note: Not for merging.

temp: check newly added test behaviour with pyjwt=2.1.0 version.

Purpose of this PR is to verify the test added in https://github.com/edx/edx-analytics-data-api/pull/468 against `pyjwt=2.1.0 version.`
It should fail because [edx-auth-backend ](https://github.com/edx/auth-backends) and `edx-drf-ext` using using old version of pyjwt

### DIFF
--- a/analytics_data_api/tests/test_jwt.py
+++ b/analytics_data_api/tests/test_jwt.py
@@ -1,0 +1,26 @@
+from time import time
+
+import jwt
+from django.test import TestCase
+
+
+class JWTSanityTest(TestCase):
+    """ Test for the JWT sanity check."""
+    def test_encode_decode(self):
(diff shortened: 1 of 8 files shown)


awais786:
It is failing
```
        if options["verify_signature"] and not algorithms:
>           raise DecodeError(
                'It is required that you pass in a value for the "algorithms" argument when calling decode().'
            )
E           jwt.exceptions.DecodeError: It is required that you pass in a value for the "algorithms" argument when calling decode().
.tox/py38/lib/python3.8/site-packages/jwt/api_jwt.py:86: DecodeError
```

PR #470, opened 2021-08-09
feat: Upgrading edx-drf-extensions which will bring pyjwt latest ver.

awais786:
Upgrading edx-drf-extensions which will bring pyjwt latest version with its breaking changes.

[Release notes](https://github.com/edx/edx-drf-extensions/releases/tag/7.0.0)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,8 +33,18 @@ tox.requirements:  ## install tox requirements
 develop: test.requirements  ## install test and dev requirements
 	pip3 install -q -r requirements/dev.txt
 
+
+COMMON_CONSTRAINTS_TXT=requirements/common_constraints.txt
+.PHONY: $(COMMON_CONSTRAINTS_TXT)
+$(COMMON_CONSTRAINTS_TXT):
+	wget -O "$(@)" https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt || touch "$(@)"
+
 upgrade: export CUSTOM_COMPILE_COMMAND=make upgrade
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -33,8 +33,16 @@ tox.requirements:  ## install tox requirements
 develop: test.requirements  ## install test and dev requirements
 	pip3 install -q -r requirements/dev.txt
 
+
+COMMON_CONSTRAINTS_TXT=requirements/common_constraints.txt
```
awais786, line 37: We want to rollout this change on few IDA's with owner's team testing/review.

That's why not removing the `pyjwt` pin from common constraint but to bring latest version on `pyjwt==2.1.0` we have to do these steps. 

On make upgrade It's download the common constraint locally and paste its content in local constraint file and then remove the `pyjwt` pin.

Zacharis278 (commented):
this makes sense to me, we probably want to coordinate with enterprise on testing this since they're the only MFE making requests.

Makefile
schenedx: I am both impressed by the creative solution, and worried that this might not be cleaned up when we actually remove the pyjwt constraint. Anyways. I accept this for now.

schenedx (approved):
👍 Let's try this out.

awais786:
@schenedx please wait. We are releasing another version of drf-extensions. I will inform you soon.

awais786:
For common constraint issue we have a ticket now for further discussion.

iamsobanjaved (approved):

awais786:
@schenedx its ready for testing now.

schenedx:
@awais786 Merged. This will go to stage automatically. Once there, I will try test and let you know here.

PR #471, opened 2021-08-11
[Feat]: Upgrade django from 2.2 to 3.2

schenedx:
The django library 2.2 LTS is going end of life next year. Move the library to django 3.2
Also, with this upgrade, moved to use edx-api-doc-tools to generate the swagger doc UI at /docs


muhammad-ammar:
@schenedx Swagger is a helpful in testing apis locally. but in this new upgrade apis in https://github.com/edx/edx-enterprise-data are missing. Please see the old and new swagger ui below

**Old Swagger UI**

analytics-data-api-old-swagger-ui

**New Swagger UI**

analytics-data-api-new-swagger-ui

Zacharis278:
can you catch me up on the plan for common_constraints? Seems like something we still wanted to include (but with overrides) based on #470. Is it's removal just temporary until that file is updated for Django 3.2?

awais786:
To avoid new migrations you need to set this `DEFAULT_AUTO_FIELD = "django.db.models.AutoField"`
https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys

mzulqarnain1 (commented):
We need a few changes in settings with this upgrade. So far we've identified at least 3 changes needed to the Django settings for repositories being updated to use Django 3.2.

- DEFAULT_HASHING_ALGORITHM = "sha1" (only needed for services, not libraries) https://docs.djangoproject.com/en/3.2/releases/3.1/#default-hashing-algorithm-settings
- DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys
- Add "django.template.context_processors.request" to "context_processors" for DjangoTemplates
https://docs.djangoproject.com/en/3.2/releases/3.1/#django-contrib-admin

schenedx:
@muhammad-ammar To include the enterprise APIs in the swagger generated by this, I need to make [this change](https://github.com/edx/edx-enterprise-data/pull/259) on edx-enterprise-data repo. Could you review that PR?

schenedx:
@awais786 Could you help answer @Zacharis278's question on [common constraint](https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt) file? In this PR, I removed all reference to the common constraint. This is because I believe the common constraint pinned library no longer need to pinned in django 3.2 upgraded repos. Is that correct?

awais786:
@schenedx `python manage.py migrate` execute this `  Applying auth.0012_alter_user_first_name_max_length... OK` migration. It is coming with django3.2. I don't think it's any concerning point.

Related PR
https://github.com/django/django/pull/12581

awais786 (approved):

PR #472, opened 2021-08-11
Feanil/fix docs

feanil:
* Update sphinx and get the docs building again.
* Also add docs building to CI so it's harder to break the docs in the future.


schenedx (approved):
👍

PR #474, opened 2021-08-13
[Fix] Pin the version of elasticsearch library to < 7.14.0

schenedx:
Apparently, 7.14.0 library is not compatible with ES 7.10 server in prod

@edx/masters-devs-cosmonauts Please help review.

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -17,6 +17,9 @@
 # django-storages version 1.9 drops support for boto storage backend.
 django-storages<1.9                 # BSD
 
+# elasticsearch library cannot be greater than 7.14.0. 7.14.0 library is not compatible with the 7.10 server in stage and prod
+elasticsearch < 7.14.0
+
 # elasticsearch-dsl depends on elasticsearch >=7.8.0,<8.0.0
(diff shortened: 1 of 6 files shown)


alangsto (approved):
👍

requirements/constraints.txt
```diff
@@ -17,6 +17,9 @@
 # django-storages version 1.9 drops support for boto storage backend.
 django-storages<1.9                 # BSD
 
+# elasticsearch library cannot be greater than 7.14.0. 7.14.0 library is not compatible with the 7.10 server in stage and prod
+elasticsearch < 7.14.0
```
matthugs, line 21: Looking at the release notes it seems like the versions of the library mirror the latest version of ES they support. We might want to pin to https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/release-notes.html#rn-7-10-0, the first version which adds support for ES7.10

schenedx: @matthugs What you observe is fair. I also think by the fact that 7.14.0 version had an explicit warning on upgrading to 7.14 would fail means we can just use this most updated version that is compatible with 7.10 ES server.

PR #484, opened 2021-09-17
fix: ES search should return all matches, not only 10

alangsto:
## [MST-1033](https://openedx.atlassian.net/browse/MST-1033)

Elasticsearch queries were only returning 10 matches, as opposed to the entire query set. I believe this change occurred during a requirements upgrade in which elasticsearch-dsl was ugraded to version 7.4.0. 

Tests did not catch this behavior change because the testing for pagination created a set of 6 learners, meaning that an ES search query matching that entire set of learners would work as expected. Once the set of learners increases above 10, the test for pagination fails. 

I have updated the relevant ES code to return all matching documents, as opposed to 10, and also updated the pagination test.

### DIFF
--- a/analytics_data_api/v0/documents.py
+++ b/analytics_data_api/v0/documents.py
@@ -160,7 +160,12 @@ class RosterEntry(Document):
             }
             for sort_policy in sort_policies
         ])
-        return search_request.execute()
+
+        # calling count will generate an additional request to ES, but is necessary for returning the entire set of data
+        total = search_request.count()
+        # by default, ES will only return the first 10 documents, we want to get all of them
+        total_search_requests = search_request[0:total]
+        return total_search_requests.execute()
 
     @classmethod
     def get_course_metadata(cls, course_id):
--- a/analytics_data_api/v0/tests/views/test_learners.py
+++ b/analytics_data_api/v0/tests/views/test_learners.py
@@ -441,34 +441,40 @@ class LearnerListTests(LearnerAPITestMixin, VerifyCourseIdMixin, TestCaseWithAut
         self.assert_learners_returned(response, expected_users)
 
     def test_pagination(self):
-        usernames = ['a', 'b', 'c', 'd', 'e']
-        self.create_learners([{'username': username, 'course_id': self.course_id} for username in usernames])
+        num_users = 16
+        learner_list = [{'username': 'test' + str(num), 'course_id': self.course_id} for num in range(num_users)]
+        self.create_learners(learner_list)
 
-        response = self._get(self.course_id, page_size=2)
+        response = self._get(self.course_id, page_size=5)
         payload = json.loads(response.content.decode('utf-8'))
         self.assertDictContainsSubset(
             {
-                'count': len(usernames),
+                'count': num_users,
                 'previous': None,
-                'next': self.expected_page_url(self.course_id, page=2, page_size=2),
-                'num_pages': 3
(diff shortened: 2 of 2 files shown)


analytics_data_api/v0/documents.py
```diff
@@ -160,6 +160,8 @@ def get_users_in_course(
             }
             for sort_policy in sort_policies
         ])
+        total = search_request.count()
```
schenedx, line 163: Can you add some comments to these two lines? Does counting the search_requests generate unnecessary calls to ES?
Why do we need to do the next line? It's very unintuitive that the `search_request` first only include the first 10, not by default all of them.

ashultz0 (approved):
This turns the call to elasticsearch into two calls which is less than ideal

But reading elasticsearch_dsl there isn't an obvious better way and the design of the library is such that I can easily believe there isn't a better way in there. And I just wrote a ticket hoping to get rid of this whole thing, so not worth further chasing at this time.

analytics_data_api/v0/documents.py
```diff
@@ -160,6 +160,8 @@ def get_users_in_course(
...
+        search_request = search_request[0:total]
```
schenedx, line 164: I'd have a new variable called `all_search_requests` or `total_search_requests`. I just don't like re-purposing same variable this way.

schenedx (approved):
Tests looks good. Gave non-blocking comments inline. Let's try it!

analytics_data_api/v0/documents.py
ashultz0: this is actually how elasticsearch_dsl wants you to work!

analytics_data_api/v0/documents.py
```diff
@@ -160,7 +160,12 @@ def get_users_in_course(
             }
             for sort_policy in sort_policies
         ])
-        return search_request.execute()
+
+        # calling count will generate an additional request to ES, but is necessary for returning the entire set of data
+        total = search_request.count()
+        # by default, ES will only return the first 10 documents, we want to get all of them
+        total_search_requests = search_request[0:total]
```
matthugs, line 167: Spent some time looking up the library to make sure slicing like this was kosher, and pagination is the documented use case for slicing like this! Nice find!

It seems like we may want to actually paginate these queries though, rather than (correct me if I'm wrong & this isn't the setup:) pulling in all results from ES and then doing pagination on the result in python code. This approach seems like for large courses it may be really really unperformant. I wouldn't hesitate to say "let's just follow up on that in a separate ticket", but I'm worried large courses which are currently limited to 10 learners might end up in a state where they can't see any since results take a long time to come back. Might be a regression for some courses. Can we test that? Are there large enough courses on staging to replicate that sort of issue?

matthugs (changes requested):

analytics_data_api/v0/documents.py
matthugs: In discussion on slack it seems like we'll test on stage with a large course.

I'd like a follow-up ticket to explore paginating the ES queries themselves, since I think that'll be valuable either way.

matthugs: Great, here: https://openedx.atlassian.net/browse/MST-1056

alangsto: To add, we will test on stage for a course with ~28,000 students. I also created https://openedx.atlassian.net/browse/MST-1056 to explore the ES pagination question.

matthugs (approved):

PR #488, opened 2021-09-17
Fix: return first 10000 matches from ES search

alangsto:
ES is currently only returning 10 documents per search, which is not ideal. If we try to increase the scope of the search to include all matching documents, endpoints using ES will error if there are more than 10,000 matches. First attempt at fixing here: https://github.com/edx/edx-analytics-data-api/pull/484

I also tried to change from using `execute` to `scan` here https://github.com/edx/edx-analytics-data-api/pull/486, but this was causing a transport error: 
`File "/edx/app/analytics_api/venvs/analytics_api/lib/python3.8/site-packages/elasticsearch/connection/base.py", line 330, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(
elasticsearch.exceptions.TransportError: TransportError(406, 'Content-Type header [application/x-www-form-urlencoded; charset=UTF-8] is not supported', 'Content-Type header [application/x-www-form-urlencoded; charset=UTF-8] is not supported')`.

For this attempt, we are simply limiting search to 10,000 documents, so that the bug is partially fixed (better than nothing, right?). There is a follow-up ticket to determine an actual fix for paginating requests to ES.

### DIFF
--- a/analytics_data_api/v0/documents.py
+++ b/analytics_data_api/v0/documents.py
@@ -160,7 +160,14 @@ class RosterEntry(Document):
             }
             for sort_policy in sort_policies
         ])
-        return search_request.execute()
+
+        # by default, ES will only return the first 10 documents, we want to get as many as possible
+        count = search_request.count()
+        if count > 10000:
+            logger.info('Cannot return more than 10,000 learners for course_id=%s', course_id)
+            count = 10000
+        total_search_requests = search_request[0:count]
+        return total_search_requests.execute()
 
     @classmethod
     def get_course_metadata(cls, course_id):
--- a/analytics_data_api/v0/tests/views/test_learners.py
+++ b/analytics_data_api/v0/tests/views/test_learners.py
@@ -441,34 +441,40 @@ class LearnerListTests(LearnerAPITestMixin, VerifyCourseIdMixin, TestCaseWithAut
         self.assert_learners_returned(response, expected_users)
 
     def test_pagination(self):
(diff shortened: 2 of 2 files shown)


analytics_data_api/v0/documents.py
```diff
@@ -160,7 +160,10 @@ def get_users_in_course(
             }
             for sort_policy in sort_policies
         ])
-        return search_request.execute()
+
+        # by default, ES will only return the first 10 documents, we want to get as many as possible
+        total_search_requests = search_request[0:10000]
```
ashultz0, line 165: I hate to add stuff now but I think we should do the count solution and do
count
log about the clamp if count > 10000 
do search request [0:count or 10000]

so we can see how much trouble is happening

ashultz0 (changes requested):

analytics_data_api/v0/documents.py
matthugs: We can do this on monday

matthugs (approved):
I would prob do `min(count, 10000)` but UNLIKE SOMEONE (nya 😛 ) I'm not gonna BLOCK on that on a FRIDAY AFTERNOON (these are jokes, I am joking)

analytics_data_api/v0/documents.py
ashultz0: fair, we should log this but you could add that later if you want

ashultz0 (approved):

matthugs (approved):
gosh even better

PR #505, opened 2021-11-05
feat: use new database for v1 api

Zacharis278:
Adds an empty v1 api that includes all of the v0 urls and by extension views. Requests made to the v1 api will have any database queries routed to a different database.

This requires a bit of trickery because the typical database router only knows about the model it is querying and nothing about the request context. This adds a request middleware step to evaluate the path for a v1 url and store the new DB parameter on the thread where it can be referenced inside the db router.

Still thinking on it but I'm not seeing a good way to unit test this.

Open to suggestions if there is a less-scary way to do this...

@edx/masters-devs-cosmonauts 

https://openedx.atlassian.net/browse/MST-1113
https://openedx.atlassian.net/browse/MST-1112

### DIFF
--- a/analytics_data_api/v1/urls.py
+++ b/analytics_data_api/v1/urls.py
@@ -1,0 +1,5 @@
+from ..v0.urls import urlpatterns as v0_urlpatterns
+
+app_name = 'v1'
+
+urlpatterns = v0_urlpatterns.copy()
--- a/analyticsdataserver/router.py
+++ b/analyticsdataserver/router.py
@@ -1,15 +1,17 @@
+import threading
+
 from django.conf import settings
 from django.utils.deprecation import MiddlewareMixin
 
-import threading
-
 request_cfg = threading.local()
 
+
 class RouterMiddleware(MiddlewareMixin):
-    def process_view(self, request, view_func, args, kwargs):
+    def process_view(self, request):
         if 'api/v1' in request.path:
             request_cfg.database = getattr(settings, 'ANALYTICS_DATABASE_V1', 'default_v1')
-    def process_response(self, request, response):
+
+    def process_response(self, request, response):  # pylint: disable=unused-argument
         if hasattr(request_cfg, 'database'):
             del request_cfg.database
         return response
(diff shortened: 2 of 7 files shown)


analyticsdataserver/router.py
```diff
@@ -7,8 +21,11 @@ def db_for_read(self, model, **hints):  # pylint: disable=unused-argument
         return self._get_database(model._meta.app_label)
 
     def _get_database(self, app_label):
-        if app_label in ('v0', 'enterprise_data'):
-            return getattr(settings, 'ANALYTICS_DATABASE', 'default')
+        if app_label in ('v1', 'v0', 'enterprise_data'):
+            if hasattr(request_cfg, 'database'):
+                return request_cfg.database
+            else:
+                return getattr(settings, 'ANALYTICS_DATABASE', 'default')
```
ashultz0: this is interesting but I actually think for the shell we want v1 to still use v0 so that there is identical v1 data rather than empty

then as we actually make v1 calls they can connect up to v1 DB in a less exciting way

Zacharis278: you have a good point we don't need this for ALL v0 urls. I think my brain got sidetracked down the wrong path here.

analytics_data_api/v1/urls.py
```diff
@@ -0,0 +1,5 @@
+from ..v0.urls import urlpatterns as v0_urlpatterns
+
+app_name = 'v1'
+
+urlpatterns = v0_urlpatterns.copy()
```
ashultz0, line 5: this shell was much less code than I anticipated 😆

ashultz0, line 5: this shell was much less code than I anticipated 😆

schenedx: I also think if we want to add new urlpatterns to v1, we can just do that by `+=` operators below this, right? That's python magic.

analyticsdataserver/settings/local.py
```diff
@@ -66,6 +74,9 @@
 
 ANALYTICS_DATABASE = 'analytics'
 ENTERPRISE_REPORTING_DB_ALIAS = 'analytics'
+# Requests to the /v1/ api will use this database.
+# To test a second database under the /v1/ routes change this to 'analytics_v1'
+ANALYTICS_DATABASE_V1 = 'analytics'
```
schenedx, line 79: Before we merge, shall we just update this to `analytics_v` so it does not give the false impression we already have the new DB populated?

schenedx (approved):
I do not see any change here that's disruptive. So, LGTM

Zacharis278:
Probably got a bit ahead of myself on this but I'll keep the branch around because we might need this code when we do implement the new database. Closing in favor of #506 which just focuses on the API.

PR #506, opened 2021-11-08
feat: v1 api urls

Zacharis278:
@edx/masters-devs-cosmonauts 

Adds a new set of V1 API URLs that will be incrementally built out to use a new data source. Currently they just fall back to the V0 views if not implemented.

Also adds configuration for a second database that will be supporting this new API. Currently, that database is unused.

https://openedx.atlassian.net/browse/MST-1113

### DIFF
--- a/analyticsdataserver/settings/base.py
+++ b/analyticsdataserver/settings/base.py
@@ -60,6 +60,15 @@ DATABASES = {
         'USER': 'api001',
         'ATOMIC_REQUESTS': False,
     },
+    'reports_v1': {
+        'ENGINE': 'django.db.backends.mysql',
+        'HOST': 'localhost',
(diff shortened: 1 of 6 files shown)


ashultz0 (approved):

analyticsdataserver/settings/base.py
```diff
@@ -370,7 +379,9 @@
 
 ########## ANALYTICS DATA API CONFIGURATION
 
-ANALYTICS_DATABASE = 'default'
+ANALYTICS_DATABASE = 'reports'
```
Zacharis278, line 382: This change matches the actual database value getting loaded from both configuration and edx-internal. Has no effect but at least makes things consistent.

PR #510, opened 2021-12-06
Add Database Routing for v1 Analytics Data API URLs

MichaelRoytman:
The new v1 URLs will be supported by the new Aurora MySQL database. In order to ensure that v1 views are backed by this database, this pull request adds custom middleware to add a database "hint" to thread-local storage when handling v1 URLs. This "hint" is used by the AnalyticsApiRouter database router to route model operations to the Aurora MySQL database instead of the original database.

JIRA: https://openedx.atlassian.net/browse/MST-1095


analyticsdataserver/router.py
```diff
@@ -1,15 +1,18 @@
...
+
 
 class AnalyticsApiRouter:
     def db_for_read(self, model, **hints):  # pylint: disable=unused-argument
         # pylint: disable=protected-access
         return self._get_database(model._meta.app_label)
 
     def _get_database(self, app_label):
-        if app_label in ('v0', 'enterprise_data'):
+        if app_label in ('v0', 'v1', 'enterprise_data'):
```
ashultz0, line 12: does this mean it could just look at app_label = v1 right here and not have to look at the thread_data?

MichaelRoytman:
There is some refactoring in this pull request. The test file `analyticsdataserver/tests.py` wasn't being run, because pytest looks for files with the `test_` prefix. I elected to break the file up into a few smaller files; `test_tests.py` just didn't seem right. A lot of the code changes are addressing this refactor and fixing tests that were broken in that file.

analyticsdataserver/router.py
MichaelRoytman: Unfortunately, I don't think so. My understanding is that we do not have any models with the `v1` app label. The app label will be the application that the model is defined in. All the models will have the `v0` label until we add models to the `v1` app.

I suppose this code change doesn't do anything, so I'm happy to remove it. I was being overly cautious in seeing this and thinking, "this may trip us up in the future if we add models to v1 and don't realize we need to change this."

analyticsdataserver/settings/base.py
```diff
@@ -221,13 +221,14 @@
...
-    'analytics_data_api.v0.middleware.ParameterValueErrorMiddleware',
-    'analytics_data_api.v0.middleware.ReportFileNotFoundErrorMiddleware',
-    'analytics_data_api.v0.middleware.CannotCreateDownloadLinkErrorMiddleware',
+    'analytics_data_api.middleware.LearnerEngagementTimelineNotFoundErrorMiddleware',
+    'analytics_data_api.middleware.LearnerNotFoundErrorMiddleware',
+    'analytics_data_api.middleware.CourseNotSpecifiedErrorMiddleware',
+    'analytics_data_api.middleware.CourseKeyMalformedErrorMiddleware',
+    'analytics_data_api.middleware.ParameterValueErrorMiddleware',
+    'analytics_data_api.middleware.ReportFileNotFoundErrorMiddleware',
+    'analytics_data_api.middleware.CannotCreateDownloadLinkErrorMiddleware',
```
MichaelRoytman, line 230: I'm just fixing the import here; nothing new.

analyticsdataserver/tests/test_clients.py
```diff
@@ -0,0 +1,95 @@
+import json
```
MichaelRoytman, line 1: This should be the same as the tests that were in `analyticsdataserver/tests.py`.

analyticsdataserver/tests/test_utils.py
```diff
@@ -0,0 +1,22 @@
+import logging
```
MichaelRoytman, line 1: This should be the same as the tests that were in `analyticsdataserver/tests.py`.

analyticsdataserver/tests/test_views.py
```diff
@@ -0,0 +1,65 @@
+import unittest.mock as mock
```
MichaelRoytman, line 1: This should be the same as the tests that were in `analyticsdataserver/tests.py`.

analyticsdataserver/tests/utils.py
```diff
@@ -0,0 +1,22 @@
+from django.contrib.auth.models import User
```
MichaelRoytman, line 1: This should be the same as the tests that were in `analyticsdataserver/tests.py`.

analyticsdataserver/router.py
ashultz0: ah I think I understand

that sounds like good caution to me

analytics_data_api/middleware.py
```diff
@@ -14,6 +16,23 @@
...
+    """
+    Add a database hint, analyticsapi_database, in the form of an attribute in thread-local storage.
+    This is used by the AnalyticsApiRouter to switch databases between the v0 and v1 views.
+    """
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if 'api/v1' in request.path:
+            thread_data.analyticsapi_database = getattr(settings, 'ANALYTICS_DATABASE_V1', None)
```
ashultz0, line 32: it'd be nice to have an else clause to always set analyticsapi_database instead of having it be a maybe

analyticsdataserver/tests/utils.py
ashultz0: these are helpful notes

another method I like is to break refactors into a separate commit so you can look at just the interesting changes or just the boring changes

ashultz0 (approved):
looks good, consider always setting the database thread local

analytics_data_api/middleware.py
schenedx: I agree with @ashultz0. We should handle the case the settings value is returning None.

analytics_data_api/tests/test_middleware.py
```diff
@@ -0,0 +1,19 @@
...
+from analytics_data_api.v0.tests.views import CourseSamples
+from analyticsdataserver.tests.utils import TestCaseWithAuthentication
+
+
+class RequestVersionMiddleware(TestCaseWithAuthentication):
+    def test_request_version_middleware_v1(self):
+        self.authenticated_get('/api/v1/courses/{}/activity'.format(
+            CourseSamples.course_ids[0]))
+
+        self.assertEqual(thread_data.analyticsapi_database, getattr(settings, 'ANALYTICS_DATABASE_V1'))
```
schenedx, line 13: Please add a test case to test the case where the setting `ANALYTICS_DATABASE_V1` is missing values

schenedx (commented):
I have the same comments as @ashultz0. Rest LGTM

analytics_data_api/middleware.py
MichaelRoytman: What should the `else` value be, in your opinion? `default`? If a Django router returns a `None` value from any of its four methods (e.g. `db_for_read`) then it automatically delegates to the `default` database. I don't want to make a decision in this middleware of what database to use in the else case. I think that the else case should be up to the router, not this middleware. I think this middleware's responsibility is to determine whether a particular database is required by the version of the API that is being used, but I don't think it should know what to select if the API is not being used. I would rather it fall back to the default behavior of the database router. What do you think?

ashultz0: I see what you're saying but to me the mental model is the thread_data as set in the middleware is only the analyticsapi_database, it's not actually making a statement about how that is used. It's just saying this DB is the analytics DB for this URL. Analytics DB is V1 if it's in the URL, otherwise it's regular analytics DB.

later your router says hey I need analytics DB and the thread_data may have info for me

I don't feel very strongly about this, I just like values that are set. So up to you.

schenedx: No strong opinion here with me either. With your explanation, this looks good for me.

MichaelRoytman: I took in your feedback.

PR #515, opened 2022-01-13
[MST-1230] Make ANALYTICS_DATABASE_V1 Testable in Devstack

MichaelRoytman:
This change makes v1 of the Analytics Data API testable in devstack.

The changes include:
* modifying the generate_fake_course_data management command to accept a "database" option
* modifying the loaddata make target to run the generate_fake_course_data management command with the analytics_v1 database
* adding a new problem_response_answer_distribution_analytics_v1 JSON file and loading it as part of the loaddata make target
* adding a development specific Django router
* modifying Django settings in the based, development, and test environment to use the correct Django routers

When users of devstack run the loaddata make target, both the analytics and the analytics_v1 databases will be loaded with dummy data. The data will be different, so users will be able to verify differences in the APIs.

### DIFF
--- a/analytics_data_api/fixtures/problem_response_answer_distribution_analytics_v1.json
+++ b/analytics_data_api/fixtures/problem_response_answer_distribution_analytics_v1.json
@@ -1,0 +1,274 @@
+[
+    {
+        "fields": {
+            "answer_value": "36.02736",
+            "correct": false,
+            "first_response_count": 0,
+            "last_response_count": 2,
+            "course_id": "EarthSciences/GP202/Spring2014",
+            "created": "2014-06-24T17:13:11",
+            "module_id": "i4x://EarthSciences/GP202/problem/d50bf059fa05468e94f61d52dd37f228",
+            "part_id": "i4x-EarthSciences-GP202-problem-d50bf059fa05468e94f61d52dd37f228_4_1",
+            "value_id": null,
+            "variant": null,
+            "problem_display_name": "Earth Science Question",
+            "question_text": "Enter your answer:"
+        },
+        "model": "v0.problemfirstlastresponseanswerdistribution",
+        "pk": 1
+    },
+    {
+        "fields": {
+            "answer_value": "33.0",
+            "correct": false,
+            "first_response_count": 0,
+            "last_response_count": 2,
+            "course_id": "EarthSciences/GP202/Spring2014",
+            "created": "2014-06-24T17:13:11",
+            "module_id": "i4x://EarthSciences/GP202/problem/d50bf059fa05468e94f61d52dd37f228",
+            "part_id": "i4x-EarthSciences-GP202-problem-d50bf059fa05468e94f61d52dd37f228_4_1",
+            "value_id": null,
+            "variant": null,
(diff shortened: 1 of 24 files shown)


analytics_data_api/fixtures/problem_response_answer_distribution_analytics_v1.json
```diff
@@ -0,0 +1,274 @@
+[
+    {
+        "fields": {
+            "answer_value": "36.02736",
+            "correct": false,
+            "first_response_count": 0,
+            "last_response_count": 2,
```
MichaelRoytman, line 7: The only changes are that I have multiplied the `last_response_count` by two.

analytics_data_api/management/commands/tests/test_generate_fake_course_data.py
```diff
@@ -13,7 +15,8 @@ def testNormalRun(self):
             'generate_fake_course_data',
             f"--num-weeks={num_weeks}",
             "--no-videos",
-            "--course-id", course_id
+            "--course-id", course_id,
+            "--database", 'analytics',
```
MichaelRoytman, line 19: I would like to run this command with the `analytics_v1` argument as well, but because the `AnalyticsModelsRouter` prevents migrations from running on the v1 database, the tables do not exist in the database. This causes failures.

analyticsdataserver/settings/test.py
```diff
@@ -14,11 +14,27 @@
         "HOST": "",
         "PORT": "",
     },
+    'analytics': {
+        'ENGINE': 'django.db.backends.sqlite3',
```
MichaelRoytman, line 18: I just realized these may need to be different. I don't think so, given that I get v1 specific database query errors but no v0 errors, but I'm not sure. I'll try to look into it.

MichaelRoytman: I can't find information supporting or refuting this. I'll just change the name for the sake of clarity, at the very least. I'll push up the change once first pass reviews are done.

analytics_data_api/management/commands/tests/test_generate_fake_course_data.py
ashultz0: this test demonstrates that you can give it a DB name and generate fake data so a run on a different DB name doesn't prove much more

analyticsdataserver/router.py
```diff
@@ -2,28 +2,112 @@
...
+    database operations to particular databases depending on the version of the API requested via the view.
+    """
     def _get_database(self, app_label):
-        if app_label in ('v0', 'v1', 'enterprise_data'):
+        if app_label in ANALYTICS_APP_LABELS:
             if hasattr(thread_data, 'analyticsapi_database'):
                 return getattr(thread_data, 'analyticsapi_database')
+            # If there is no analyticsapi_database database hint, then fall back to next router.
+            return None
+        return None
```
ashultz0, line 73: could collapse this a little to be more obviously as simple as it is

if analytics app and hasattr
return getattr
return none

schenedx (approved):
👍

ashultz0 (approved):

PR #516, opened 2022-01-13
no enrollments age for v1

ashultz0:
We're not implementing age for v1, so take out the enrollments age URL

Which, since we were just copying the existing URLs, means put in all the URLs except this one instead of making a straight copy of v0

MST-1241


schenedx (approved):
👍

Zacharis278 (approved):

PR #518, opened 2022-01-18
fix: strip unnecessary expensive ordering

ashultz0:
Having more columns in the django official sort means we sort by them all the time. We do not index by these fields, usually we only index by course id and date which is the obvious data organization. The data API further processes these, in the case of gender and mode squashing different entries to a single date row, and Insights sorts them locally.

We can't entirely ditch sorting here without adding it in python because in processing course enrollment groupby(date, course_id) is usually used and that requires the initial list to be sorted by those keys as well. 

related to MST-1305


schenedx (approved):
👍

PR #519, opened 2022-01-19
use explicit sort in python instead of sorting in DB

ashultz0:
This PR is a bit of a performance experiment. I don't think it will blow up insights, it may help significantly, it may not.

Sorting in mysql is more efficient using indices and we're seeing the production mysql use an index without date which means it then sorts using a less efficient method. Sometimes it writes data to disk to do the sort, so that just blows up everything. Behavior is better on the read only DB than competing on the write DB, but still not always where it should be.

All this data is in memory in the API so we can make python do the sort. Conceptually I'm also not a fan of models having an order separated from your query just because it causes this sort of invisible database load.

Not all models need sorting, only the ones that use groupby, so there's asymmetry between lines removed and added.

Location code is left alone for the moment because it is not causing trouble and there is more weird stuff going on in location sort and resort than I wanted to tackle for this experiment. I also left the meta/program models alone today.

### DIFF
--- a/analytics_data_api/v0/models.py
+++ b/analytics_data_api/v0/models.py
@@ -24,7 +24,6 @@ class CourseActivityWeekly(BaseCourseModel):
     class Meta(BaseCourseModel.Meta):
         db_table = 'course_activity'
         index_together = [['course_id', 'activity_type']]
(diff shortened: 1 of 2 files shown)


schenedx (approved):
👍

alangsto (approved):

PR #520, opened 2022-01-24
fix: update health check endpoint to ensure all dbs are operational

schenedx:
[MST-1231](https://openedx.atlassian.net/browse/MST-1231)
Update the health check endpoint to check all important databases: default and (reports OR reports_v1).
If any of the dbs are not accessible, the overall status of the service will be `UNAVAILABLE`.

@edx/masters-devs-cosmonauts Please review

### DIFF
--- a/analyticsdataserver/settings/test.py
+++ b/analyticsdataserver/settings/test.py
@@ -33,7 +33,7 @@ DATABASES = {
 }
 
 ANALYTICS_DATABASE = 'analytics'
-ANALYTICS_DATABASE_V1 = 'analytics_v1'
+ANALYTICS_DATABASE_V1 = None
 ENTERPRISE_REPORTING_DB_ALIAS = 'default'
 
 # Silence elasticsearch during tests
--- a/analyticsdataserver/views.py
+++ b/analyticsdataserver/views.py
@@ -71,36 +71,60 @@ class HealthView(APIView):
 
    """
     permission_classes = (permissions.AllowAny,)
+    STATUS_OK = 'OK'
+    STATUS_UNAVAILABLE = 'UNAVAILABLE'
 
-    def get(self, request, *args, **kwargs):  # pylint: disable=unused-argument
-        OK = 'OK'
-        UNAVAILABLE = 'UNAVAILABLE'
-
-        overall_status = UNAVAILABLE
-        db_conn_status = UNAVAILABLE
-
+    def _get_connection_status(self, connection_name):
+        """
+        With the passed in database name, try to make connection to the database.
+        If database is able to connect successfully, return 'OK' status.
+        Otherwise, return 'UNAVAILABLE' status
+        """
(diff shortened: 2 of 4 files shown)


analyticsdataserver/views.py
```diff
@@ -71,36 +71,55 @@ class HealthView(APIView):
...
+        overall_status = self.STATUS_UNAVAILABLE
+
+        # First try the default database.
+        # The default database hosts user and auth info. Without it, no API would function
+
+        default_db_status = self._get_connection_status('default')
+        analytics_db_status = self._get_connection_status(
+            getattr(settings, 'ANALYTICS_DATABASE')
+        )
+        analytics_v1_db_status = self._get_connection_status(
```
ashultz0, line 106: it should probably only do this if the system cares about the v1 DB, i.e. settings.DATA_API_V1_ENABLED = True

schenedx: @ashultz0 There is no such setting `DATA_API_V1_ENABLED`. That setting is instead defined in the [Insights repo](https://github.com/openedx/edx-analytics-dashboard/search?q=DATA_API_V1_ENABLED)

Zacharis278: I imagine we don't want this to fail for existing systems w/ no v1 DB. Should we maybe have a different 'v1' healthcheck endpoint if we need to rely on the error response code?

analyticsdataserver/settings/test.py
```diff
@@ -33,7 +33,7 @@
 }
 
 ANALYTICS_DATABASE = 'analytics'
-ANALYTICS_DATABASE_V1 = 'analytics_v1'
+ANALYTICS_DATABASE_V1 = None
```
ashultz0, line 36: this might fail the router tests, not sure

analyticsdataserver/views.py
```diff
@@ -71,36 +71,55 @@ class HealthView(APIView):
...
+
+        # First try the default database.
+        # The default database hosts user and auth info. Without it, no API would function
+
+        default_db_status = self._get_connection_status('default')
+
+        analytics_db_name = getattr(settings, 'ANALYTICS_DATABASE_V1')
+        if not analytics_db_name:
+            # If the v1 db is not defined, we use the default analytics db name setting instead
+            analytics_db_name = getattr(settings, 'ANALYTICS_DATABASE')
```
ashultz0, line 107: while we are transitioning we need both to be defined to be healthy

and there are enterprise tables in the old analytics db that we aren't planning to move. So I think this should be if v1 check it, and always check v0

analyticsdataserver/settings/test.py
schenedx: I checked and it does not

ashultz0 (approved):

Zacharis278 (approved):

PR #524, opened 2022-02-01
fix: ignore health check for newrelic reporting

ashultz0:
performance of the health check is not particularly interesting and
the ELB hits it enought to obscure reporting on stuff we care about


schenedx (approved):
👍

matthugs:
applied https://github.com/edx/terraform/pull/4299 which may obviate this

PR #525, opened 2022-02-02
fix user activity casing

ashultz0:
we've got trash case data in the user activity tables because of mixed case sensitivity across the pipeline

handle that in the API layer


ashultz0:
better to do this in the data source, closing

PR #527, opened 2022-02-02
feat: update fake data commands to generate stage data

alangsto:
MST-1284. A new management command is needed to generate partial, incremental data for stage. Much of this command could be based off of the existing fake data generation command, so the common pieces were abstracted to a base command. As opposed to nuking all existing data, the stage command will incrementally build off of the most recent date. Stage data will also not include any age data.


analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,339 @@
...
+            logger.info("Generating new daily course enrollment data...")
+
+        logger.info("Generating new daily course enrollment data...")
+
+        # Create new data
+        daily_total = 1500
+        date = start_date
+
+        enrollment_objs = models.CourseEnrollmentModeDaily.objects.using(database).filter(course_id=course_id)
+        cumulative_count = enrollment_objs.latest().cumulative_count \
```
alangsto, line 110: start incrementing cumulative count, as opposed to starting from 0 every time we run the command

analytics_data_api/management/commands/generate_stage_course_data.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def get_start_date(self, database):
+        enrollments = models.CourseEnrollmentDaily.objects.using(database)
+        # default to ten weeks of data if no data exists
+        start_date = datetime.datetime.now() - datetime.timedelta(weeks=4)
+        if enrollments:
+            latest_date = enrollments.latest().date
+            # want to return latest date + some time difference to avoid overlap
+            start_date = datetime.datetime.combine(
+                latest_date + datetime.timedelta(days=1), datetime.datetime.min.time()
```
alangsto, line 33: Start adding data a day after our last enrollment record date, that way we can continually add data each week

alangsto:
I didn't change the tests for the `generate_fake_course_data` command tests, can add additional tests to up the code coverage if we think it's needed

analytics_data_api/management/commands/generate_stage_course_data.py
```diff
@@ -0,0 +1,64 @@
...
+            '--database',
+            action='store',
+            dest='database',
+            default='default',
+            help='Database in which to generate fake date',
+        )
+
+    def get_start_date(self, database):
+        enrollments = models.CourseEnrollmentDaily.objects.using(database)
+        # default to ten weeks of data if no data exists
```
ashultz0, line 27: comment says 10, constant says 4

analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,339 @@
+# pylint: disable=line-too-long,invalid-name
+"""
+Abstract base management command meant to be called by other commands that generate fake data
```
ashultz0, line 3: still combing through but I think this could be just free functions instead of a command object 

it's not managing state, all the generate commands already nicely take everything they need. "self" is just a decoration

analytics_data_api/management/commands/generate_data.py
ashultz0: good to point out, add a comment

analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,339 @@
...
+# http://stackoverflow.com/a/3590105
+def constrained_sum_sample_pos(num_values, total):
+    """Return a randomly chosen list of n positive integers summing to total.
+    Each such list is equally likely to occur."""
+
+    dividers = sorted(random.sample(range(1, total), num_values - 1))
+    return [a - b for a, b in zip(dividers + [total], [0] + dividers)]
+
+
+def get_count(start):
```
ashultz0, line 34: this name is not that helpful but I tried a bit to think of a better one and can't

at least change start to base or min?

analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,339 @@
...
+            ratios = [n / 1000.0 for n in constrained_sum_sample_pos(len(birth_years_range), 1000)]
+            birth_years = dict(list(zip(birth_years_range, ratios)))
+
+        if delete_data:
+            # Delete existing data
+            for model in [models.CourseEnrollmentDaily,
+                          models.CourseEnrollmentModeDaily,
+                          models.CourseEnrollmentByGender,
+                          models.CourseEnrollmentByEducation,
+                          models.CourseEnrollmentByBirthYear,
```
ashultz0, line 95: if there is no birth year table in a db what does this delete do?

alangsto: It doesn't complain, just tested on my local environment

analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,339 @@
...
+        date = start_date
+
+        enrollment_objs = models.CourseEnrollmentModeDaily.objects.using(database).filter(course_id=course_id)
+        cumulative_count = enrollment_objs.latest().cumulative_count \
+            if (use_current_cumulative and enrollment_objs) \
+            else 0
+
+        progress = tqdm(total=(end_date - date).days + 2)
+        while date <= end_date:
+            daily_total = get_count(daily_total)
```
ashultz0, line 116: this chunk could be profitably split out to increase readability, it's generating one day

analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,339 @@
...
+            'UNKNOWN': 0.01
+        }
+
+        enrollment_mode_ratios = {
+            'audit': 0.15,
+            'credit': 0.15,
+            'honor': 0.25,
+            'professional': 0.10,
+            'verified': 0.35
+        }
```
ashultz0, line 81: all of these ratios and the birth year ratios (which we could make all the time) could live outside the function to increase function readability

analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,339 @@
...
+                course_id=course_id, catalog_course_title='Demo Course', catalog_course='Demo_Course',
+                start_time=timezone.now() - datetime.timedelta(weeks=6),
+                end_time=timezone.now() + datetime.timedelta(weeks=10),
+                pacing_type='self_paced', availability='Starting Soon', enrollment_mode=mode, count=count,
+                cumulative_count=cumulative_count, count_change_7_days=random.randint(-50, 50),
+                passing_users=int(cumulative_count * pass_rate))
+
+        progress.update(1)
+        progress.close()
+        logger.info("Done!")
```
ashultz0, line 169: "done generating enrollment data"? or don't bother logging at all

analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,339 @@
...
+            models.CourseActivityWeekly.objects.using(database).all().delete()
+            logger.info("Deleted all weekly course activity.")
+
+        logger.info("Generating new weekly course activity data...")
+
+        progress = tqdm(total=math.ceil((end_date - start).days / 7.0))
+        while start < end_date:
+            active_students = random.randint(100, 4000)
+            # End date should occur on Saturday at 23:59:59
+            end = start + datetime.timedelta(weeks=1)
```
ashultz0, line 188: the new snowflake data generates weeks as monday to sunday
it also generates start and end at midnight, not 23:59:59
ie. 2019-07-07 00:00:00.000 to 2019-07-14 00:00:00.0000
so this needs a what is a week arg
or maybe make it run wednesday to wednesday to be Deliberately Fake
and in either case let's not use 59:59

analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,339 @@
...
+            low_ceil = random.random() * max_value * 0.5
+            models.ModuleEngagementMetricRanges.objects.using(database).create(
+                course_id=course_id, start_date=start_date, end_date=end_date, metric=event,
+                range_type='low', low_value=0, high_value=low_ceil)
+            high_floor = random.random() * max_value * 0.5 + low_ceil
+            models.ModuleEngagementMetricRanges.objects.using(database).create(
+                course_id=course_id, start_date=start_date, end_date=end_date, metric=event,
+                range_type='high', low_value=high_floor, high_value=max_value)
+
+    def generate_tags_distribution_data(self, course_id, database):
```
ashultz0, line 260: I don't even know if we can see this tags stuff anywhere in insights currently but this is not the time to drop it

analytics_data_api/management/commands/generate_stage_course_data.py
```diff
@@ -0,0 +1,64 @@
...
+    def handle(self, *args, **options):
+        course_ids = ['course-v1:edX+DemoX+Demo_Course']
+        insights_course_ids = ['course-v1:edX+Insights+Stage', 'course-v1:edX+Insights+Fake+Data']
+        course_ids += insights_course_ids
+        usernames = ['ed_xavier', 'xander_x', 'alice_bob']
+        program_title = 'edX Insights'
+        database = options['database']
+
+        logger.info("Option to generate videos with ids pulled from the LMS is disabled, using fake video ids...")
+        video_ids = self.fake_video_ids_fallback()
```
ashultz0, line 46: nice

analytics_data_api/management/commands/generate_stage_course_data.py
```diff
@@ -0,0 +1,64 @@
...
+                course_id, start_date, end_date, database,
+                delete_data=False, add_birth_year=False, use_current_cumulative=True
+            )
+            self.generate_all_video_data(course_id, video_ids, database)
+            for username in usernames:
+                self.generate_learner_engagement_data(course_id, username, start_date, end_date, database)
+            self.generate_learner_engagement_range_data(course_id, start_date.date(), end_date.date(), database)
+            self.generate_tags_distribution_data(course_id, database)
+
+        self.generate_program_data(insights_course_ids, program_title, database)
```
ashultz0, line 64: is this supposed to be insights_course_ids rather than course_ids?

analytics_data_api/management/commands/tests/test_generate_stage_course_data.py
```diff
@@ -0,0 +1,108 @@
...
+@set_databases
+class GenerateStageCourseDataTests(TestCase):
+
+    def test_run_with_no_preexisting_data(self):
+        """
+        Test that data is generated for the past four weeks
+        """
+        course_ids = [
+            'course-v1:edX+DemoX+Demo_Course', 'course-v1:edX+Insights+Stage', 'course-v1:edX+Insights+Fake+Data'
+        ]
```
ashultz0, line 20: make these accessible from the command so they don't need to be duplicated here

analytics_data_api/management/commands/tests/test_generate_stage_course_data.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def test_run_with_preexisting_data(self):
+        """
+        Test that data is generated for the time between preexisting data and now,
+        e.g. if data was added a week ago, should only add a weeks worth of data.
+        No previous data should be deleted.
+        """
+        course_ids = [
+            'course-v1:edX+DemoX+Demo_Course', 'course-v1:edX+Insights+Stage', 'course-v1:edX+Insights+Fake+Data'
+        ]
```
ashultz0, line 62: here again

I'm imagining updating them in the command and then updating these copies

analytics_data_api/management/commands/generate_stage_course_data.py
alangsto: Yes, wanted to only have a subset of the courses be in a program so we could check that filtering by program worked.

analytics_data_api/management/commands/generate_data.py
alangsto: Taking the wednesday-wednesday approach

analytics_data_api/management/commands/generate_data.py
```diff
@@ -0,0 +1,349 @@
...
+        models.CourseActivityWeekly.objects.using(database).all().delete()
+        logger.info("Deleted all weekly course activity.")
+
+    logger.info("Generating new weekly course activity data...")
+
+    progress = tqdm(total=math.ceil((end_date - start).days / 7.0))
+    while start < end_date:
+        active_students = random.randint(100, 4000)
+        # End date should occur on Wednesday at 00:00:00
+        end = (start + datetime.timedelta(weeks=1)).replace(hour=0, minute=0, second=0, microsecond=0)
```
ashultz0, line 193: shouldn't need to do the replace here, midnight + 1 week is still midnight

ugh except for daylight savings, how does that work with python timedelta addition
although that still doesn't work here if you lose an hour and end up on tuesday and then reset to start of tuesday
unless the datetime is UTC? which doesn't have daylight savings. In that case it's ok.

really you want to be working in dates not microsecond precision timedelta, so you can say wednesday + 7 days like a human would. But that's something that has to be done across the codebase.

alangsto: Yeah, that makes sense. Updated to just add the seven days.

PR #529, opened 2022-02-14
chore: update pip and pip-tools versions to fix version incompatibility

ashultz0:
also does a general upgrade as a side effect


ashultz0:
recording extra steps for pip upgrade here for convenience:

> pip install --upgrade pip;pip install --upgrade pip-tools;pip-compile -v --allow-unsafe --rebuild -o requirements/pip.txt requirements/pip.in;pip-compile --upgrade -o requirements/pip_tools.txt requirements/pip_tools.in;make upgrade

schenedx (approved):
👍

PR #535, opened 2022-03-01
chore: remove django-countries constraint

alangsto:
A fix has been released for django-countries 7.3, so we can remove that constraint.


schenedx (approved):
👍

PR #538, opened 2022-03-15
feat: upgrade requirements to catch enterprise data 4.2.0

ashultz0:
there's only one requirement we need today, but it's an excuse to generally upgrade


ashultz0:
closed in favor of the next PR

PR #540, opened 2022-03-16
chore: upgrade enterprise-data version

alangsto:
Update enterprise data to the latest version, which contains a fix to handle rate limiting when making calls to the LMS to retrieve enterprise customer details.


schenedx (approved):
👍

PR #543, opened 2022-04-11
feat: all db migrations handled by one router

Zacharis278:
Unifies database router logic so that when migrate is run in any environment it produces the same result. There was some weirdness even before we picked this project up that would cause the development environment to differ from production. The output of migrate now matches production exactly.

Unfortunately, I was unsuccessful at making this work without at least one bit of janky code regarding the enterprise migrations. The crux of our issues here are old migrations related to roles in the enterprise app that used to run in the default database, presumably prior to the use of prefect. This app was moved to a different database, and the role based models recreated in a new app.  The new app `enterprise_data_roles` exists in the default database alongside django auth because it needs a foreign key to auth_user, which you cannot do cross-DB in mysql. Thing is, the database migrations for the old role models still exist and if the `enterprise_data` migrations are run in a separate database (where it now belongs) everything explodes.

My recommendation is we squash those useless and problematic migrations, and then remove the condition in here regarding 'DJANGO_AUTH_MODELS' in the future to clean this up. 

Followup Ticket: https://openedx.atlassian.net/browse/MST-1462

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [pycodestyle]
-ignore=E501,E741
+ignore=W504,E501,E741
 max_line_length=119
 exclude=settings,analyticsdataserver/wsgi.py
 
--- a/analytics_data_api/tests/test_middleware.py
+++ b/analytics_data_api/tests/test_middleware.py
@@ -1,5 +1,3 @@
-from unittest import mock
-
 from django.conf import settings
 from django.test import override_settings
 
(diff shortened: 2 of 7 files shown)


setup.cfg
```diff
@@ -1,5 +1,5 @@
 [pycodestyle]
-ignore=E501,E741
```
Zacharis278, line 2: W504 and W503 are mutually exclusive and both enabled meaning it's impossible to write a multiline if statement. We follow W503 in other repos so let's do that here.

ashultz0 (approved):
good cleanup in here in a tough situation, step 2 of the earlier router changes

schenedx (commented):
👍

PR #547, opened 2022-04-20
feat: format answer distribution fields

Zacharis278:
Answer values in the old data pipeline went through formatting so they could be presentable in the insights UI. The new pipeline has no such formatting so these values need to be cleaned after being fetched from the database.

Also we need to do some additional processing to undo replacements made due two problematic escape sequences in snowflake.

https://openedx.atlassian.net/browse/MST-1454

### DIFF
--- a/analytics_data_api/v0/serializers.py
+++ b/analytics_data_api/v0/serializers.py
@@ -1,5 +1,8 @@
+import json
+import re
 from collections import OrderedDict
 
+import html5lib
 from django.conf import settings
 from rest_framework import pagination, serializers
 from rest_framework.response import Response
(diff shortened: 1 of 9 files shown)


analytics_data_api/v0/serializers.py
```diff
@@ -96,6 +99,108 @@ class Meta:
...
+        exists between elements (e.g. table elements).
+        """
+        try:
+            root = html5lib.parse(markup)
+            text_list = []
+            for val in self._get_text_from_element(root):
+                text_list.extend(val.strip().split(' '))
+            text = u' '.join(text_list)
+        except Exception:  # pylint: disable=broad-except
+            text = markup.strip()
```
Zacharis278, line 162: I can't seem to create data that would actually cause this function to throw an exception but this is in the old pipeline so I figure we keep this to be safe. Can't hit w/ coverage tho

alangsto (commented):
This looks good to me! I'm curious how much time it adds to the existing answer_distribution data, since that data was already cleaned during the EMR pipeline. I think the serializer is probably the easiest solution but just want to make sure performance doesn't take a huge hit if the Open edX community is still using this.

Zacharis278:
@alangsto since we pull this one problem at a time it doesn't need to run through a lot of fields. No noticeable issues locally but going to do some further testing on stage w/ more data.

analytics_data_api/v0/serializers.py
ashultz0: the only possible place for exceptions is html5lib.parse, I can't find a clear description of what exceptions that would throw on a quick reading. If it wants to deal with HTML in the world there is probably no possible malformed HTML that will make it throw an exception.

I'd replace the content here - if something is so horrible that it caused html5lib to choke, we probably should blank it rather than carry it along to transmit over the web to a browser

ashultz0 (approved):
one minor thought only

analytics_data_api/v0/serializers.py
Zacharis278: I can get behind just blanking this because you're right, if we end up here something probably gone wrong.

PR #548, opened 2022-04-21
feat: add default throttling to analytics-data-api

alangsto:
We currently don't have any rate limiting in place for the analytics-data-api. This PR defines default throttling rates for users and services, and uses a custom throttling class to account for service users. 

Service users can be updated via the `ANALYTICS_API_SERVICE_USERNAMES` setting.

This type of throttling is similar to what's done in course-discovery.


schenedx (approved):
👍

PR #550, opened 2022-04-25
remove boto, tox, ordered_set constraints

ashultz0:
Again was unable to mess with pylint constraints without drawing up a tangle of required changes. If we want to develop the data API more we'll want to make those but it doesn't seem worth doing presently.


schenedx (approved):
👍

PR #553, opened 2022-05-09
fix: new blackslash quote slug value

Zacharis278:
pairs with: https://github.com/edx/warehouse-transforms/pull/3133


alangsto (approved):
👍

PR #554, opened 2022-05-10
Align course_id and usage_id Styles Among Test Data and With LMS

MichaelRoytman:
The fixtures `problem_response_answer_distribution_analytics.json` and `problem_response_answer_distribution_analytics_v1.json` represent `answer_distribution` data that is loaded into the `answer_distribution` table when using the `loaddata` Make target. These fixtures use the old style course IDs for course_id in the form `edX/DemoX/Demo_Course` and the old style for the usage ID module_id in the form `i4x://edX/DemoX.1/problem/05d289c5ad3d47d48a77622c4a81ec36`.

The data that is set up as part of the `generate_fake_course_data` management command, also run as part of the `loaddata` Make target, defaults to using a new style course ID in the form `course-v1:edX+DemoX+Demo_Course`.

I don't believe that this is an issue for the Analytics Data API on its own. However, the development environment for Insights (the Analytics Data Dashboard) is closely coupled to the Analytics Data API. In devstack, Insights uses new style IDs, so requests to the problems endpoint in the Analytics Data API fail, because there are no problems in the database with the new style course ID. This pull request fixes the course ID style to match.

Furthermore, in order to render the Performance tab and its constituent views, Insights leverages the course API in the LMS to return the structure of the course when rendering the performance views. This means that Insights looks for `module_ids` in the `answer_distribution` table match the usage ids of problems in the edX Demo Course in devstack. Currently, the `module_ids` in the fixtures do not match any problem id in the edX Demo Course. This results in there being no data displayed for the performance views. Although this is very brittle, the easy fix here is to ensure that the `module_ids` and `part_ids` match usage ids of problems in the edX Demo Course.


alangsto (approved):
👍

Zacharis278 (approved):

PR #564, opened 2022-06-28
chore: upgrade requirements

ashultz0:
Automated upgrade failed but local upgrade works when I expected it to fail in the same way, might as well take the upgrade results.

I suspect this upgrade PR will get us past the problematic pip versions.


schenedx (approved):
👍

PR #567, opened 2022-07-07
fix: choose v1 database for problems if it exists

alangsto:
The insights performance data tab is not displaying data correctly after our pipeline cutover on May 20th. 

Looking in splunk, every time you try and access data from the performance tab, an error like `Resource https://prod-edx-analyticsapi-internal.edx.org/api/v1/courses/<course_id>/problems/ was not found on the API server.`. I found that the problems endpoint is still trying to hit the old database. 

We can either prioritize the new DB if it exists (although this will impact the v0 problems endpoint), or we can create a new view for the v1 endpoint (this option would make cleanup a bit messier).


ashultz0 (approved):
my sadness is intense

ashultz0:
I think the tests may run all on one DB unfortunately 😢

analytics_data_api/v0/views/courses.py
```diff
@@ -665,7 +665,11 @@ def get_queryset(self):
 GROUP BY module_id;
         """
 
-        connection = connections[settings.ANALYTICS_DATABASE]
+        # If possible, get database from router
+        router_db = router.db_for_read(ProblemResponseAnswerDistribution)
+        database = router_db if router_db else settings.ANALYTICS_DATABASE
```
ashultz0, line 670: I think the router has to return something in django or you keep going to deeper routers until you get the default so this may not be needed

ashultz0 (approved):

Zacharis278 (approved):
nice, this is definitely better 👍

ashultz0:
this should also preserve our ability to do v=0 to see the old ones and do a quick comparison (although one obvious feature should be that the new one has updated recently)

analytics_data_api/v0/views/courses.py
alangsto: Oh yep, I think you're right. I will remove that piece.

PR #577, opened 2022-08-23
ADR for no PII in analytics API

ashultz0:
includes tiny fix to the other decision doc since I was next to it


alangsto (approved):
Makes sense to me 👍

